Issue #1, opened 2014-08-24
Add Tests


jpadilla:
We should be able to reuse the tests from [ember-simple-auth-devise](https://github.com/simplabs/ember-simple-auth/tree/master/packages/ember-simple-auth-devise/tests/simple-auth-devise)

jpadilla:
Just released [v0.4.0](https://github.com/jpadilla/ember-cli-simple-auth-token/releases/tag/v0.4.0).

PR #12, opened 2014-10-02
Extracts the method that builds the token

kurko:
I need to have quotes around my token (e.g `Token token="myusertoken"`). I feel it's inconvenient to have to rewrite the entire `authorize` method just for that.

On a side note, I think someone should try to come up with the basic structure for tests. There are no tests for this repo.


jpadilla:
@kurko thanks for this. I agree with you, tests are my #1 priority right now. I'm taking a look at the new [Addon CLI options](http://www.ember-cli.com/#developing-addons-and-blueprints), once I get that working, we'll have a structure for tests that should work well.

In the meantime any work for adding tests to this is definitely welcome.

PR #24, opened 2015-01-17
Adds JWT Authenticator with refresh functionality

erichaus:


jpadilla:
@erichonkanen hey mind check out the failing build?

https://travis-ci.org/jpadilla/ember-cli-simple-auth-token/builds/47388220

erichaus:
@jpadilla there we go...

erichonkanen:
@jpadilla, in doing more testing I discovered that if you do a page refresh, it doesn't trigger the automatic token refresh... where would I look to implement that?

erichaus:
Added restore ability so that automatic refresh begins with valid token on page refresh.. Now what's left seems a little daunting but writing more tests? It might take me a bit to read up on how to write them..

jpadilla:
@erichonkanen definitely needs tests. There's tons of things going on in this PR so tests will help me understand and check everything is working correctly, great work!

erichaus:
@jpadilla definitely agree! I'll be working on tests and let you know when it's updated..

erichaus:
@jpadilla made progress on first tests for restore method, check them out and lmk if they on the right path. They cover each of the possible outcomes within the method... https://github.com/erichonkanen/ember-cli-simple-auth-token/blob/561d31a3d6a1fd46b8275bd0bd881ebf6ba8f5f2/tests/unit/authenticators/jwt-test.js

Also I don't see the travis ci build in this PR anymore, not sure if it needs to be resubmitted or whatnot? I'll aim to have this done by friday!

jpadilla:
@erichonkanen nice, glad you got them working! Looking good. I think you might need to update your fork. There were some changes that happened a while back that are causing conflicts, that might be why builds aren't running.

erichonkanen:
spent some time (a lot!) trying to get sinon.spy(Ember.run, 'later') to work with the authenticate tests to no avail.. ended up trying import QUnit and abstracting to testStart() method. It works but runs it for every test. If you have better suggestion lmk..

https://github.com/erichonkanen/ember-cli-simple-auth-token/blob/master/tests/unit/authenticators/jwt-test.js#L24-L30

erichaus:
@jpadilla good news and bad news.. the good news is I finished all the tests so hypothetically this can be  reviewed now... the bad news is that I discovered that opening a 2nd tab with the app instantly boots me out of both tabs. So I'll have to do some digging into that, if you have any ideas around that lmk! otherwise tell me if anything outside of that needs changes... thanks!

erichaus:
Fixed multiple tab logout issue, added one more test, reviewed test coverage and all seems well!

jpadilla:
@erichonkanen what was causing the "multiple tab logout issue" ?

erichaus:
@jpadilla For some reason I had made a barely noticeable change and added parenthesis to reject in call to _this.refreshAccessToken. That made it fire the reject always when restore was called and thus boot me out on page refresh or loading new tab...

```
_this.refreshAccessToken(data.token).then(function(data){
     resolve(data);
}, reject());
```

Changed `reject()` back to `reject`

of

```
restore: function(data){
    var _this = this;
    return new Ember.RSVP.Promise(function(resolve, reject){
      var now = (new Date()).getTime(),
        expiresAt = _this.resolveTime(data.expiresAt);
      if(!Ember.isEmpty(data.expiresAt) && !Ember.isEmpty(data.token) && data.expiresAt > now){
        if(_this.refreshAccessTokens){
          _this.refreshAccessToken(data.token).then(function(data){
            resolve(data);
          }, reject);
        }else{
          reject();
        }
      }else{
        if(Ember.isEmpty(data.token)){
          reject();
        }else{
          var tokenData = _this.getTokenData({'token': data.token}),
            tokenExpiresAt = tokenData[_this.tokenExpireName];
          expiresAt = _this.resolveTime(tokenExpiresAt);
          _this.scheduleAccessTokenRefresh(expiresAt, data.token);
          resolve(data);
        }
      }
    });
  },
```

jpadilla:
@erichonkanen I don't think `timeFactor` defaulting to `1` is a good idea, perhaps that was just for testing?

This'll take me more than I thought to review. I think it'll be a good idea to build a demo app that uses with a demo backend server.

erichaus:
@jpadilla so the purpose of timeFactor was to normalize converting between e.g. python's default seconds to javasripts default milliseconds... I'm not sure which should be the default, in my case since Im using python it would make sense to have it be "1000". Maybe there is a better way to handle that conversion?

Also, yes I recommend setting up a demo app. In fact I have one up you could log into but its private so pm me.. Ive tested it with 2 separate apps and it works great. Its helpful to set the token refresh on python side to e.g. 15 seconds for testing that it works

erichaus:
shoot I think you're right in that time factor should be default 1000 since this package is being developed against django rest jwt...

I had just set it in my project config to     timeFactor: 1000,

jpadilla:
@erichonkanen whenever you can hit me up with access wherever you've got this working on.

erichaus:
@jpadilla sweet! everything good? thanks a ton for all the help in this!

jpadilla:
@erichonkanen I noticed that you weren't using `tokenPropertyName`, so in a few places I think you're assuming that `token` will be in the root of the response object.

Also, I think `getTokenData` should receive the token string, instead of a response object.

erichaus:
@jpadilla ok gotcha, ill take a look and make some updates hopefully tomorrow if not early this week.. thx for the review

PR #25, opened 2015-01-21
Add support for custom headers

leandrocp:
Add an optional customHeader configuration, wich will send custom headers on every authentication request.

Just like http://emberjs.com/api/data/classes/DS.RESTAdapter.html#toc_headers-customization

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,5 +95,6 @@ ENV['simple-auth-token'] = {
   tokenPropertyName: 'token',
   authorizationPrefix: 'Bearer ',
   authorizationHeaderName: 'Authorization',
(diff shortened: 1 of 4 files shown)


jpadilla:
@leandrocp looks good to me. How about calling it `headers` instead of `customHeader` just like how it's done in `DS.RESTAdapter`?

leandrocp:
@jpadilla that´s fine to me. I´ll update the pull request.

leandrocp:
@jpadilla done!

jpadilla:
@leandrocp Thanks!

PR #27, opened 2015-02-17
Extract token from properties if tokenPropertyName indicates nested values

pentagon:
This target cases when tokenPropertyName looks like `user.token` i.e. whole user object gets stored as a result of authentication.

### DIFF
--- a/addon/authenticators/token.js
+++ b/addon/authenticators/token.js
@@ -100,7 +100,7 @@ export default Base.extend({
   restore: function(properties) {
     var _this = this;
     return new Ember.RSVP.Promise(function(resolve, reject) {
-      if (!Ember.isEmpty(properties[_this.tokenPropertyName])) {
+      if (!Ember.isEmpty(_this.extractTokenFromProperties(properties))) {
         resolve(properties);
       } else {
         reject();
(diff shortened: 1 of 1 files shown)


jpadilla:
@pentagon could we make `extractTokenFromProperties` a `utils` function? It can be more generic by passing the object and the string(`user.token`).

Also, while we're at it we can make sure it works on multiple levels, like `user.data.token`? This SO [answer](http://stackoverflow.com/a/8052100/4054812) might be worth a look.

A simple test might also come in handy.

pentagon:
Yes, we can do everything we're capable to as well. I'm pretty sure that you knows better how that piece of code should be organized. I've just shown a way how we could address quite general case of spotted misbehaving.

jpadilla:
@pentagon creating an Ember.Object from `properties` should work. Trying it out.

jpadilla:
@pentagon thanks for the initial work on this.

Managed to get it done in c2efe97524dc7b513fbbfc86544d5623c1762e3a

pentagon:
thank you very much!

jpadilla:
@pentagon released v0.5.1 with this.

PR #29, opened 2015-02-23
Changed getTokenData to accept token string

erichaus:


jpadilla:
@erichonkanen I think we need to use `tokenPropertyName` to avoid assuming `response.token`. And do something like I did on c2efe97524dc7b513fbbfc86544d5623c1762e3a.

erichaus:
@jpadilla ah that's what you meant.. I see, updated to reflect that, lmk if that's looking better?

erichaus:
Hey,

So I tried converting the code in restore() to use the Ember.Object but I'm getting broken tests etc saying "undefined is not a function" for dataObject.get('expiresAt'), am I doing something wrong here?

```
  restore: function(data) {
    var _this = this,
      dataObject = Ember.Object.create(data);

    return new Ember.RSVP.Promise(function(resolve, reject) {
      var now = (new Date()).getTime(),
        expiresAt = _this.resolveTime(dataObject.get('expiresAt'));

      if (!Ember.isEmpty(data.get('expiresAt')) && !Ember.isEmpty(data.get(_this.tokenPropertyName)) && expiresAt > now) {
        if (_this.refreshAccessTokens) {
          _this.refreshAccessToken(data.get(_this.tokenPropertyName)).then(function(data) {
            resolve(data);
          }, reject);
        } else {
          reject();
        }
      } else {
        if (Ember.isEmpty(data.get(_this.tokenPropertyName))) {
          reject();
        } else {
          var tokenData = _this.getTokenData(data.get(_this.tokenPropertyName)),
            tokenExpiresAt = tokenData[_this.tokenExpireName];

          _this.scheduleAccessTokenRefresh(tokenExpiresAt, data.get(_this.tokenPropertyName));

          resolve(data);
        }
      }
    });
  },
```

jpadilla:
@erichonkanen hard to tell, could you inspect data and dataObject?

erichaus:
@jpadilla fixed it... tests passing, lmk what you think

jpadilla:
@erichonkanen looks good to me, thanks!

Issue #32, opened 2015-03-16
JWT automatic refresh, configurable interval?


jdhoek:
The automatic refresh for JSON Web Tokens works, but it is cutting it kind of close by refreshing at the absolute last second that the previous token is still valid. Just a bit of lag could cause the old token to reach the API server in an expired state.

I would like to refresh the token a little while before the old one expires. Can this margin be made a configurable parameter?

erichaus:
@jdhoek seems like a good feature... when I was building that I was using a [django backend jwt package](https://github.com/GetBlimp/django-rest-framework-jwt) that has a jwt_leeway setting which I had set to 3 seconds.. It makes sense to have that in this package as well

PR #33, opened 2015-03-16
Add a margin to the automatic JWT refresh

jdhoek:
By configuring `refreshMargin` the JWT gets refreshed that many seconds
before the current token expires.

PR for #32.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,10 +87,24 @@ export default Ember.Controller.extend(LoginControllerMixin, {
 
 Please note, the JWT authenticator will decode a token and look for the
 expiration time found by looking up the token[Config.tokenExpireName]. It then
-calculates the difference between the current time and the token expire time
-to determine when to make the next automatic token refresh request.
+calculates the difference between the current time and the token expire time —
+from which the *refreshLeeway* is subtracted — to determine when to make the
+next automatic token refresh request.
 
-For example, your decoded token might look like this:
+For example, with the following configuration:
+
+```
+  ENV['simple-auth'] = {
+    authorizer: 'simple-auth-authorizer:token'
+  };
+  ENV['simple-auth-token'] = {
+    refreshAccessTokens: true,
+    timeFactor: 1,
+    refreshLeeway: 300 // Refresh the token 5 minutes (300s) before it expires.
+  };
+```
+
+Your decoded token might look like this:
 
 ```
 token = {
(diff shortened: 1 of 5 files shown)


jdhoek:
I've had a go at implementing the feature I suggested in #32 by adding a configurable parameter that defaults to 0. It might actually make sense to default to some value (10 seconds? A minute? 5 minutes?), but I'm just getting my feet wet with JSON Web Tokens, so I might be mistaken.

validkeys:
+1 on this one. All of my refresh requests keep getting cancelled as the JWT is expired by the time the refresh makes the requests.

Great work @jdhoek

jpadilla:
I definitely thought this would have been an issue. @erichonkanen haven't you run into this?

This looks pretty good as is to me, not too sure about the actual name for the setting `refreshMargin`, definitely not against it. We use `leeway` in [PyJWT](https://github.com/jpadilla/pyjwt).

erichaus:
@jpadilla I have not run into this issue in any of the projects Ive used this with. Two of these projects are deployed to heroku currently in an ember-cli client/django backend setup with proxy. 

The only thing I can think of for me is that I have JWT_LEEWAY set to 3 seconds on the django jwt side. 

That said it seems logical to have something like this.

cheers

erichaus:
@validkeys have you tried setting a leeway on the server side? that is the most likely reason you ran into this and I didn't

jdhoek:
@jpadilla 
_leeway_ does sound better, I'll push a commit renaming this to _refreshLeeway_.

I'm implementing a Java API that serves up and refreshes JWT. I considered giving the timeout some leeway server-side, but that would encourage an API where the client sends expired or very nearly expired tokens to the server. That doesn't feel quite right in terms of a clean API, although for all practical purposes both approaches should work equally fine.

jdhoek:
Thanks for the :+1: vote of confidence.

@jpadilla 
I've updated the documentation in `README.md` as well, please have a look if all is in order there.

validkeys:
@erichonkanen I didn't but I just tried it out and that works well for me. Having that hint as part of the documentation would be great for JWT noobs (like me) (Best Practice would be to include a leeway in your JWT's decode method server side).

validkeys:
@erichonkanen @jpadilla put a quick note into the docs: jpadilla/ember-cli-simple-auth-token/pull/34

jdhoek:
Thanks!

PR #34, opened 2015-03-17
Add server-side "leeway" hint to documentation

validkeys:
Adding server-side leeway hint and example ruby code

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,17 +103,7 @@ token = {
 In this case the token expire name is using the default `exp` as set by the
 `Config.tokenExpireName` property.
 
-An automatic token refresh request would be sent out at token[Config.tokenExpireName] - now(). Best practice with regards to token refreshing is to also set a "leeway" on your server-side Jwt decode method. An example in ruby:
-
-```ruby
-require 'ruby-jwt'
-
-def decode_jwt
-  token = request.headers['Authorization'].split(' ').last
-  JWT.decode(token, 'secret', true, {leeway: 10}) # adds 10 seconds of leeway to decode the jwt post-expiry
-end
-
-```
+An automatic token refresh request would be sent out at token[Config.tokenExpireName] - now(). A good practice with regards to token refreshing is to also set a "leeway" when decoding JSON Web Tokens in the server-side. Some libraries like [PyJWT](https://github.com/jpadilla/pyjwt) and [ruby-jwt](https://github.com/progrium/ruby-jwt) already support this.
 
 ## The Authorizer
 
--- a/README.md
+++ b/README.md
@@ -103,7 +103,17 @@ token = {
 In this case the token expire name is using the default `exp` as set by the
 `Config.tokenExpireName` property.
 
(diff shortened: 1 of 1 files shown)


jpadilla:
I'd probably rephrase that a bit and would remove the example code and just link to some libraries that do support this.

> An automatic token refresh request would be sent out at `token[Config.tokenExpireName] - now()`. A good practice with regards to token refreshing is to also set a "leeway" when decoding JSON Web Tokens in the server-side. Some libraries like [PyJWT](https://github.com/jpadilla/pyjwt) and [ruby-jwt](https://github.com/progrium/ruby-jwt) already support this.

validkeys:
Good points. Updated the PR. Thanks @jpadilla

jpadilla:
@validkeys mind updating this PR so I can merge?

jdhoek:
I agree with the extra note regarding server-side leeway, but judging from the [JWT spec](https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#section-3), I'm not convinced that it is meant to address the case of refreshing a token at the exact moment of expiration.

The relevant part:

> The "exp" (expiration time) claim identifies the expiration time on
>    or after which the JWT MUST NOT be accepted for processing.  The
>    processing of the "exp" claim requires that the current date/time
>    MUST be before the expiration date/time listed in the "exp" claim.
>    Implementers MAY provide for some small leeway, usually no more than
>    a few minutes, to account for clock skew.

This server-side leeway is to prevent a little bit of clock-skew on the client or server from expiring a token too soon. If both server and client are presumed to be NTP-synched devices (most consumer devices are these days), that leeway would be quite small.

From an API standpoint, a client should never depend on this server-side leeway, as it is exclusively  meant to address clock skew. With #33 the client-side leeway can be configured to suit the target usage scenario (e.g. a webapp meant for mobile use in areas with spotty network coverage would probably want to refresh much sooner than a webapp meant exclusively for use from intranet desktop computers).

I would recommend seperating the two cases in the documentation (client-side leeway for token refreshing and server-side leeway for addressing clock skew). I'm not sure if it is helpful to recommend refreshing tokens with an expired JWT as good practice if (from a security standpoint) it is not allowed according to the spec. What do you think @validkeys ?

@jpadilla 
Following the JWT spec, it might even be better to include a minor client-side leeway by default, as the spec forbids accepting tokens that expire at exactly _now_. Currently the refresh happens **at** `exp`, if I'm not mistaken.

I hope I don't sound too pedantic. :smile: My main concern is that a lot of users from a library such as this depend on its documentation for making the correct security assumptions. Most developers will never read the JWT spec itself.

validkeys:
@jpadilla hey sorry, I've been out of town. I'll take care of this today.

@jdhoek I can definitely see your point on either. My perspective is that, as you pointed out, there are many different situations that can come up. I think optionally supporting bother client-side and server-side leeway gives the lib the most flexibility to meet all of these different potential use-cases -- and I don't think it causes bloat to support both. That's my what I think at least

validkeys:
@jpadilla ok updated

PR #38, opened 2015-03-26
Allow for strings to be used as token data

knownasilya:
Pretty much `JSON.parse` chokes if you give it a string (valid json).

### DIFF
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -12,7 +12,7 @@ import TokenAuthenticator from './token';
 
   @class JWT
   @namespace SimpleAuth.Authenticators
-  @module simple-auth-token/authenticators/jwt
+  @module simple-auth-token/authenticgators/jwt
   @extends TokenAuthenticator
 */
 export default TokenAuthenticator.extend({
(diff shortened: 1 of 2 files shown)


jpadilla:
@knownasilya any chance of a small test?

knownasilya:
@jpadilla yeah let me figure out how to create a fake token with data.

knownasilya:
@jpadilla does that look good?

jpadilla:
Yup. Quick question, what's your use case for a string like that? Probably mean it's not a JSON Web Token, so why use this authenticator?

knownasilya:
@jpadilla it is a jwt, just the data I put into it is a string, instead of an object. This is the most basic jwt token, without exp or anything.

jpadilla:
Interesting. The [spec](https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#section-4) defines the JWT Claim as:

> The JWT Claims Set represents a JSON object whose members are the claims conveyed by the JWT.

So why not just send a JSON object with just one key and your value?

knownasilya:
@jpadilla I probably will do that eventually :+1: just wanted to fix this so it can handle this use case, since https://github.com/auth0/node-jsonwebtoken allows a string as data.

jpadilla:
@knownasilya definitely valid, just wanted to make sure. Thanks again!

PR #41, opened 2015-04-01
don't restore expired tokens

calvinmetcalf:
I noticed a bug then when I logged into a site, then navigated away, then after the token expired navigating back the the now expired token will be accepted and the site will act like you are logged in until you try to do something with the token at which point you get logged out.

This pull also cleans up the logic around restoring tokens, uses real error objects and cleans up some of the tests to make sure they are actually testing what they intended to test.

Previous restore logic
1.  if token exists, expiresAt exists, and data.expiresAt is greater then now
   1. if refreshAccessTokens refresh the token, then return the data
   2. reject with undefined
2. else
   1. if token is empty reject with undefined
   2. else schedule a refresh based on the time in the token and return the data

the problem was with an expired token which took path 2, 2 

new logic
1. If token is empty, reject with an error
2.  if data.expiresAt is empty
   1. set expiresAt from the token
   2. if expiresAt is still empty resolve with the data (token doesn't expire).
3. If expiresAt is NaN (shouldn't be possible any more) reject with error
4. If expiresAt is greater then now (token is not expired)
   1. If time until expiration is more then the leeway
      1. if refreshAccessTokens then schedule a refresh
      2. either way resolve data
   2. else if refreshAccessTokens
      1. refresh the access token and then return the data
   3. else reject the token (we are within the leeway so force the user to get a new token instead of one which is going to be useless very soon).
5. reject as token has expired.

### DIFF
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -101,34 +101,46 @@ export default TokenAuthenticator.extend({
   restore: function(data) {
     var _this = this,
       dataObject = Ember.Object.create(data);
-    
+
     return new Ember.RSVP.Promise(function(resolve, reject) {
       var now = (new Date()).getTime(),
         expiresAt = _this.resolveTime(dataObject.get('expiresAt')),
         token = dataObject.get(_this.tokenPropertyName);
+      if (Ember.isEmpty(token)) {
+        return reject(new Error('empty token'));
+      }
 
-      if (!Ember.isEmpty(expiresAt) && !Ember.isEmpty(token) && expiresAt > now) {
-        if (_this.refreshAccessTokens) {
-          _this.refreshAccessToken(data.token).then(function(data) {
-            resolve(data);
-          }, reject);
+       if (Ember.isEmpty(expiresAt)) {
+        // Fetch the expire time from the token data since `expiresAt`
+        // wasn't included in the data object that was passed in.
+        var tokenData = _this.getTokenData(data[_this.tokenPropertyName]);
+        expiresAt = _this.resolveTime(tokenData[_this.tokenExpireName]);
+        if (Ember.isEmpty(expiresAt)) {
+          return resolve(data);
+        }
+
+      }
+      if (expiresAt !== expiresAt) {
+        return reject(new Error('invalid expiration'));
+      }
+      if (expiresAt > now) {
+        var wait = expiresAt - now - (_this.refreshLeeway * 1000);
+        if (wait > 0) {
+          if (_this.refreshAccessTokens) {
+            _this.scheduleAccessTokenRefresh(expiresAt, token);
+          }
+          resolve(data);
+        } else if (_this.refreshAccessTokens) {
+          resolve(_this.refreshAccessToken(data.token).then(function () {
+            return data;
+          }));
         } else {
-          reject();
+          reject(new Error('unable to refresh token'));
         }
       } else {
-        if (Ember.isEmpty(token)) {
-          reject();
-        } else {
-          // Fetch the expire time from the token data since `expiresAt`
-          // wasn't included in the data object that was passed in.
-          var tokenData = _this.getTokenData(data[_this.tokenPropertyName]),
-            tokenExpiresAt = tokenData[_this.tokenExpireName];
-
-          _this.scheduleAccessTokenRefresh(tokenExpiresAt, token);
-
-          resolve(data);
-        }
+        reject(new Error('token is expired'));
       }
+
     });
   },
 
@@ -179,7 +191,7 @@ export default TokenAuthenticator.extend({
     Schedules a token refresh request to be sent to the backend after a calculated
     `wait` time has passed.
 
-    If both `token` and `expiresAt` are non-empty, and `expiresAt` minus the optional 
+    If both `token` and `expiresAt` are non-empty, and `expiresAt` minus the optional
     refres leeway is greater than the calculated `now`, the token refresh will be scheduled
     through Ember.run.later.
 
@@ -255,7 +267,7 @@ export default TokenAuthenticator.extend({
   */
   getTokenData: function(token) {
     var tokenData = atob(token.split('.')[1]);
-    
+
     try {
       return JSON.parse(tokenData);
     } catch (e) {
@@ -306,6 +318,9 @@ export default TokenAuthenticator.extend({
     @private
   */
   resolveTime: function(time) {
+    if (Ember.isEmpty(time)) {
+      return time;
+    }
     return new Date(time * this.timeFactor).getTime();
   },
 });
--- a/tests/unit/authenticators/jwt-test.js
+++ b/tests/unit/authenticators/jwt-test.js
@@ -215,7 +215,6 @@ test('#restore resolves when `expiresAt` is greater than `now`', function() {
       // Check that Ember.run.later was not called.
       deepEqual(Ember.run.later.getCall(0), null);
     }).catch(function (err) {
-      console.log(err.message);
       ok(false);
     });
   });
--- a/tests/unit/authenticators/jwt-test.js
+++ b/tests/unit/authenticators/jwt-test.js
@@ -60,6 +60,7 @@ test('assigns timeFactor from the configuration object', function() {
 });
 
 test('#restore resolves when the data includes `token` and `expiresAt`', function() {
+  expect(1);
   var jwt = JWT.create(),
     expiresAt = (new Date()).getTime() + 60000;
 
@@ -71,7 +72,7 @@ test('#restore resolves when the data includes `token` and `expiresAt`', functio
 
   var data = {};
   data[jwt.tokenPropertyName] = token;
-  data[jwt.tokenExpireName] = expiresAt;
+  data.expiresAt = expiresAt;
 
   App.server.respondWith('POST', '/api-token-refresh/', [
     201, {
@@ -83,12 +84,15 @@ test('#restore resolves when the data includes `token` and `expiresAt`', functio
   Ember.run(function() {
     App.authenticator.restore(data).then(function(content) {
       // Check that the resolved data matches the init data.
-      deepEqual(content, data);
+      equal(JSON.stringify(content), JSON.stringify(data));
+    }, function () {
+      ok(false);
     });
   });
 });
 
 test('#restore resolves when the data includes `token` and excludes `expiresAt`', function() {
+  expect(1);
   var jwt = JWT.create(),
     expiresAt = (new Date()).getTime() + 60000;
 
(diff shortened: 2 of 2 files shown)


addon/authenticators/jwt.js
```diff
@@ -306,6 +318,9 @@ export default TokenAuthenticator.extend({
     @private
   */
   resolveTime: function(time) {
+    if (Ember.isEmpty(time)) {
```
calvinmetcalf, line 321: previously if the expiresAt is undefined then previously NaN would be returned

tests/unit/authenticators/jwt-test.js
```diff
@@ -83,12 +84,15 @@ test('#restore resolves when the data includes `token` and `expiresAt`', functio
   Ember.run(function() {
     App.authenticator.restore(data).then(function(content) {
       // Check that the resolved data matches the init data.
-      deepEqual(content, data);
+      equal(JSON.stringify(content), JSON.stringify(data));
```
calvinmetcalf, line 87: this gives a better error message then deepEquals

tests/unit/authenticators/jwt-test.js
```diff
@@ -71,7 +72,7 @@ test('#restore resolves when the data includes `token` and `expiresAt`', functio
 
   var data = {};
   data[jwt.tokenPropertyName] = token;
-  data[jwt.tokenExpireName] = expiresAt;
+  data.expiresAt = expiresAt;
```
calvinmetcalf, line 75: since the method never looks for `data[jwt.tokenExpireName]` this is equivalent to not setting the value so assumed it to be an error, if it just shouldn't be there I can update this

tests/unit/authenticators/jwt-test.js
```diff
@@ -110,15 +113,17 @@ test('#restore resolves when the data includes `token` and excludes `expiresAt`'
...
-      deepEqual(content, data);
+      equal(JSON.stringify(content), JSON.stringify(data), 'Check that the resolved data matches the init data');
+    }).catch(function () {
+      ok(false);
     });
   });
 });
 
-test('#restore rejects when `refreshAccessTokens` is false', function() {
+test('#restore rejects when `refreshAccessTokens` is false and token is expired', function() {
```
calvinmetcalf, line 123: if we have a valid token this shouldn't reject

tests/unit/authenticators/jwt-test.js
```diff
@@ -261,8 +275,9 @@ test('#restore does not schedule a token refresh when `refreshAccessTokens` is f
 });
 
 test('#restore does not schedule a token refresh when `expiresAt` < now.', function() {
+  expect(1);
   var jwt = JWT.create(),
-    expiresAt = (new Date()).getTime() + 60000;
+    expiresAt = (new Date()).getTime() - 10;
```
calvinmetcalf, line 280: expires at was not < now previously

tests/unit/authenticators/jwt-test.js
```diff
@@ -300,7 +316,7 @@ test('#restore does not schedule a token refresh when `expiresAt` - `refreshLeew
   App.authenticator.refreshLeeway = 120;
 
   Ember.run(function() {
-    App.authenticator.restore(data).then(function() {
+    App.authenticator.restore(data).catch(function() {
```
calvinmetcalf, line 319: if expiresAt - refreshLeeway < now we shouldn't be able to restore

tests/unit/authenticators/jwt-test.js
```diff
@@ -496,15 +514,15 @@ test('#refreshAccessToken triggers the `sessionDataUpdated` event on successful
 
 test('#getTokenData returns correct data', function() {
   var jwt = JWT.create();
-  
+
```
calvinmetcalf, line 517: sorry about the white space but my editor automatically removes trailing white space

jpadilla:
@calvinmetcalf This is one of the greatest and most well explained pull request I've ever seen, thanks! @erichonkanen since you worked with this closely, it would be great if you checked it out as well.

Unrelated, but saw you're in Boston. I'll be heading down there next Wednesday. Will probably go to the Boston Ember.js Meetup on Thursday, but looking for things to do/places to work from Friday and/or Saturday.

calvinmetcalf:
Thanks you, you might have talked to @knownasilya my coworker before. I'm mobile right now but I can give you some suggestions money.

addon/authenticators/jwt.js
```diff
@@ -101,34 +101,46 @@ export default TokenAuthenticator.extend({
...
-        } else {
-          // Fetch the expire time from the token data since `expiresAt`
-          // wasn't included in the data object that was passed in.
-          var tokenData = _this.getTokenData(data[_this.tokenPropertyName]),
-            tokenExpiresAt = tokenData[_this.tokenExpireName];
-
-          _this.scheduleAccessTokenRefresh(tokenExpiresAt, token);
 
+      }
+      if (expiresAt !== expiresAt) {
```
erichaus, line 123: hey @calvinmetcalf, is this line testing for NaN? just curious of the logic

addon/authenticators/jwt.js
```diff
@@ -101,34 +101,46 @@ export default TokenAuthenticator.extend({
...
+      if (expiresAt > now) {
+        var wait = expiresAt - now - (_this.refreshLeeway * 1000);
+        if (wait > 0) {
+          if (_this.refreshAccessTokens) {
+            _this.scheduleAccessTokenRefresh(expiresAt, token);
+          }
           resolve(data);
+        } else if (_this.refreshAccessTokens) {
+          resolve(_this.refreshAccessToken(data.token).then(function () {
+            return data;
```
erichaus, line 135: does this need `return resolve(data)` or just `resolve(data)`? just curious why this line and L132, and L119 differ?

addon/authenticators/jwt.js
```diff
@@ -101,34 +101,46 @@ export default TokenAuthenticator.extend({
...
-          // Fetch the expire time from the token data since `expiresAt`
-          // wasn't included in the data object that was passed in.
-          var tokenData = _this.getTokenData(data[_this.tokenPropertyName]),
-            tokenExpiresAt = tokenData[_this.tokenExpireName];
-
-          _this.scheduleAccessTokenRefresh(tokenExpiresAt, token);
 
+      }
+      if (expiresAt !== expiresAt) {
+        return reject(new Error('invalid expiration'));
```
erichonkanen: TIL about javascript Error object... nicely done

erichaus:
@calvinmetcalf thanks for submitting this it looks good... have you tested it with a backend etc? just curious. Tomorrow I'm going to do a quick run through with my django backend... I'm also wondering what might have been the issue with navigating back to a page you had logged out of and being able to access it(?) I hadn't seen that but helps me learn if I can identify what exactly it was?

addon/authenticators/jwt.js
calvinmetcalf, line 135: inside of a promise callback returning a value causes the promise to resolve into that value, so the fact we passed it to resolve in 134 is engouth

addon/authenticators/jwt.js
calvinmetcalf, line 123: yup

calvinmetcalf:
yes I've tested it against a node backend

erichaus:
@calvinmetcalf what was the issue? was the token being cached? I did not have that issue so I'm wondering

calvinmetcalf:
@erichonkanen if you navigated to the app and you had a token that had expired while you where off the page then the restore function would test  [if token exists, expiresAt exists, and data.expiresAt is greater then now](https://github.com/jpadilla/ember-cli-simple-auth-token/blob/5b92e2ad933240cd29ecbdabff145154d3679696/addon/authenticators/jwt.js#L110) it would fail that because data.expiresAt is less then now, then [since token was there](https://github.com/jpadilla/ember-cli-simple-auth-token/blob/5b92e2ad933240cd29ecbdabff145154d3679696/addon/authenticators/jwt.js#L119) it would [schedule a refresh and resolve](https://github.com/jpadilla/ember-cli-simple-auth-token/blob/5b92e2ad933240cd29ecbdabff145154d3679696/addon/authenticators/jwt.js#L127)

But scheduling a refresh wouldn't do anything because [wait is less then 0](https://github.com/jpadilla/ember-cli-simple-auth-token/blob/5b92e2ad933240cd29ecbdabff145154d3679696/addon/authenticators/jwt.js#L196) meaning the app would act like you were logged in until you tried to test it.

erichaus:
I see.. the thing I didn't get is how you could log in if you were off the page but that makes sense

calvinmetcalf:
you already logged in, then you left the page is how it comes up

erichaus:
@calvinmetcalf I noticed that now when the auth-token-refresh request is sent out, it is also refreshing/requesting all other data endpoints (twice on my demo app) on the page whereas previously it would only send out the auth-token-refresh request... any ideas on why?

http://i.imgur.com/SynPpEH.png

calvinmetcalf:
is this when you are restoring the app or what exactly is the context ?

erichaus:
I log into the app and navigate to localhost:8001/people/ (that page makes call to /api/v1/users/) and then after 15 seconds (made short expire time) it sends the refresh request as well as repeating the api call to /users/. Functionally it works but I'm wondering if it should only be making the auth refresh request or what would cause it to also repeat other requests?

erichaus:
I also just noticed another issue which may have existed prior and not been caught but I just noticed it now. I opened 2 tabs at once and left both tabs to browse elsewhere, returned after a minute or so and saw they both had stopped refreshing and had 400 signature expired error

calvinmetcalf:
> I also just noticed another issue which may have existed prior and not been caught but I just noticed it now. I opened 2 tabs and once and left both tabs, returned after a minute or so and saw they both had stopped refreshing and had 400 signature expired error

yeah there is likely a race condition in there 

> I log into the app and navigate to localhost:8001/people/ (that page makes call to /api/v1/users/) and then after 15 seconds (made short expire time) it sends the refresh request as well as repeating the api call to /users/. Functionally it works but I'm wondering if it should only be making the auth refresh request or what would cause that?

not totally sure didn't touch code related to that

erichaus:
the repeated data requests don't happen in my version I use on a heroku hosted app which was a commit or 2 behind so something did change but I'll look at those commits as well.. I don't think it should be making repeated data calls

calvinmetcalf:
@erichonkanen that could be the new leeway stuff on the refresh

erichaus:
@calvinmetcalf ok I tested with the main master branch and it indeed is having some type of issue with multiple browsers open that causes immediate race/refresh loop... I just went through various commits and can't determine what's causing it but I'll start a new issue for that (strange bc when I first did this feature I had same issue but resolved it and this appears to be something different). I am curious if you notice the same issue? I want to make sure its not something local to me.. other than that nice work on this

jpadilla:
@erichonkanen could you perhaps open an issue with more details on what you're experiencing, your current settings, etc. ?

PR #44, opened 2015-04-12
Fixes jwt authenticator restore method to schedule access token refresh after page reload

erichaus:
The restore method was not functioning after a page reload due to inconsistency with stored token expire name... Updated it to use the tokenExpireName throughout and restored functionality

### DIFF
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -206,7 +206,7 @@ export default TokenAuthenticator.extend({
 
       if (!Ember.isEmpty(token) && !Ember.isEmpty(expiresAt) && wait > 0) {
         Ember.run.cancel(this._refreshTokenTimeout);
-        
+
         delete this._refreshTokenTimeout;
 
         if (!Ember.testing) {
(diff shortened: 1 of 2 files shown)


addon/authenticators/jwt.js
```diff
@@ -127,11 +126,11 @@ export default TokenAuthenticator.extend({
         var wait = expiresAt - now - (_this.refreshLeeway * 1000);
         if (wait > 0) {
           if (_this.refreshAccessTokens) {
-            _this.scheduleAccessTokenRefresh(expiresAt, token);
+            _this.scheduleAccessTokenRefresh(dataObject.get(_this.tokenExpireName), token);
           }
           resolve(data);
         } else if (_this.refreshAccessTokens) {
-          resolve(_this.refreshAccessToken(data.token).then(function () {
+            resolve(_this.refreshAccessToken(data.token).then(function () {
```
jpadilla: Are we sure this should be `data.token`? Not immediately sure.

erichaus:
@jpadilla updated that to use tokenPropertyName which seems more correct :)

PR #49, opened 2015-04-27
Fix authorizer cannot build a token.

wuron:
Hi there.
I had an issue with token authorizer was unable to build token for authorizing the request and it appeared that it was trying to get it from wrong property. So here's a pull-request fixing this.

You can check that token is actually stored in `session.secure` in [`session#setup`](https://github.com/simplabs/ember-simple-auth/blob/master/packages/ember-simple-auth/lib/simple-auth/session.js#L246) method of ember-simple-auth.

### DIFF
--- a/addon/authorizers/token.js
+++ b/addon/authorizers/token.js
@@ -94,6 +94,6 @@ export default Base.extend({
     @return {String}
   */
   buildToken: function() {
-    return this.get('session.' + this.tokenPropertyName);
+    return this.get('session.secure.' + this.tokenPropertyName);
   }
 });
--- a/tests/unit/authorizers/token-test.js
+++ b/tests/unit/authorizers/token-test.js
@@ -40,7 +40,7 @@ test('#authorize when authenticated adds token to request', function() {
   expect(3);
 
   App.authorizer.set('session.isAuthenticated', true);
-  App.authorizer.set('session.token', 'secret token!');
+  App.authorizer.set('session.secure', {token: 'secret token!'});
   App.authorizer.authorize(App.request, {});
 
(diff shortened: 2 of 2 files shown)


jpadilla:
@wuron I think this changed in a newer version of `ember-simple-auth`. Which are you running? We might also need to update our requirement.

wuron:
Yeah, checked `ember-simple-auth` history. It was changed in `0.8.0-beta.1`
I'm using latest `0.8.0-beta.2`, that's the version you get if you follow [installation instructions](https://github.com/simplabs/ember-cli-simple-auth#installation).

jpadilla:
There are other couple of things that need to be done for best compatibility with that 0.8.0 release. You're free to take this on or I will once a non beta release comes out for it.

Did you install ember-simple-auth before this package by yourself? This package installs the expected version itself when you do:

```
npm install --save-dev ember-cli-simple-auth
ember generate ember-cli-simple-auth
```

devinus:
Is there a way to monkey patch this for now?

devinus:
For anybody wanting this now, try this:

``` js
import Token from 'simple-auth-token/authorizers/token';

export function initialize(/* container, application */) {
  Token.reopen({
    buildToken() {
      return this.get('session.secure.' + this.tokenPropertyName);
    }
  });
}

export default {
  name: 'authenticator',
  before: 'simple-auth-token',
  initialize: initialize
};
```

jpadilla:
Apparently I had fixed this a while back but never pushed changes. Could you try with latest master which also uses `0.8.0-beta.2`?

wuron:
Latest release works great, so I guess this can be closed.

PR #51, opened 2015-05-05
Add basic access authentication

julientherier:
Hi,

I use a backend based on Express.js and Passport.js for authentication.
So I found useful to add basic authentication support to fit [how Passport implements that process](http://passportjs.org/guide/basic-digest/).

It's basically based on a config var called `basicAuthentication` sets to `false` by default (in this case it uses your previous code).
If `basicAuthentication` is set to `true`, then the username and password are combined and encoded in the Authorization header using a GET request and default contentType.

Also some tests added.

Hope it fits with your requirements and thanks for your great addon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ For the Token authenticator:
 // config/environment.js
 ENV['simple-auth-token'] = {
   serverTokenEndpoint: '/api-token-auth/',
+  basicAuthentication: false,
   identificationField: 'username',
   passwordField: 'password',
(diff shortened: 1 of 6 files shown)


jpadilla:
This is great, but I'm a bit hesitant to include in this package. Seems a bit too specific and you can always just subclass authenticators, change `makeRequest`, and use that in your app. Let's leave this around for a little bit to see if anyone else jumps in.

PR #52, opened 2015-05-06
Token moved to secure object inside session

BryanHunt:
ember-simple-auth made a change to the session that moved the token into a sub object called "secure"


BryanHunt:
I see this PR is superseded by Issue #49 - closing.

PR #54, opened 2015-05-10
Fix jwt token refresh

tibotiber:
Hi! It seems that the automatic JWT token refresh was not working when setting a `tokenPropertyName` different from `token`. I've fixed it with this tiny commit. Hope it can help. ;)


jpadilla:
@tibotiber thanks!

PR #66, opened 2015-06-14
update time factor to 1000

ghost:


ghost:
for some reason when I run ember test everything passes

engwan:
I think this is wrong:

```
timeFactor: 1000  // example - set to "1" to convert incoming seconds to seconds.
```

The factor is used to convert everything to milliseconds, so it should be something like:

```
timeFactor: 1000  // example - set to "1" to keep incoming milliseconds as is
```

jpadilla:
@threeaccents perhaps try rebasing this with the latest upstream master branch?

PR #69, opened 2015-06-16
Fix restore method and ability to successfully page refresh

erichaus:
Opening this PR for ongoing feedback while trying to debug what is causing page refresh to break in latest 0.7.1 release.

### DIFF
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -123,13 +123,7 @@ export default TokenAuthenticator.extend({
         return reject(new Error('invalid expiration'));
       }
       if (expiresAt > now) {
-        var wait = expiresAt - now - (_this.refreshLeeway * 1000);
-        if (wait > 0) {
-          if (_this.refreshAccessTokens) {
-            _this.scheduleAccessTokenRefresh(dataObject.get(_this.tokenExpireName), token);
-          }
-          resolve(data);
-        } else if (_this.refreshAccessTokens) {
+        if (_this.refreshAccessTokens) {
             resolve(_this.refreshAccessToken(token).then(function () {
             return data;
           }));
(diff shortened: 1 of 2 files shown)


erichaus:
Not sure why the tests are failing?

```
The command "ember try $EMBER_TRY_SCENARIO test" exited with 1.
cache.2
store build cache
0.48snothing changed, not updating cache
Done. Your build exited with 1.
```

addon/authenticators/jwt.js
```diff
@@ -123,13 +123,7 @@ export default TokenAuthenticator.extend({
         return reject(new Error('invalid expiration'));
       }
       if (expiresAt > now) {
-        var wait = expiresAt - now - (_this.refreshLeeway * 1000);
-        if (wait > 0) {
```
erichaus, line 127: Removed this logic because the restore method should always immediately refresh a token and log the user in rather than schedule a token refresh for a later date.

Correct me if wrong on that though...

jpadilla:
@erichonkanen if you scroll up they are like three tests failing.

jpadilla:
@erichonkanen could you rebase this?

jrjohnson:
@erichonkanen - sorry to ping you on this, but I would love to see this merged and fixed and I think it just needs a rebase.

erichaus:
@jrjohnson this does not resolve the issue, the restore functionality is still broken and I've been trying to figure out how to fix it!

jrjohnson:
@erichonkanen Attempting to help debug this and I ran into something which may be why this is passing for some and not others.  
https://github.com/jpadilla/ember-cli-simple-auth-token/pull/69/files#diff-b3b19e97b14851cfabb1e6558abfc6daL106 uses javascripts getTime which returns a milisecond time stamp.  I was assuming a second time stamp.  This causes the straight integer comparison to fail at https://github.com/jpadilla/ember-cli-simple-auth-token/pull/69/files#diff-b3b19e97b14851cfabb1e6558abfc6daL125  It looks like the README docs ask for a second stamp as well.  You could change that comparison to use something like MomentJs to parse the value or change the values of `now` to the right factor with `Math.floor(Date.now() / 1000)`

OR just assume a milisecond factor timestamp will be passed in and change the docs accordingly.

erichaus:
@jrjohnson there has been an issue brought up regarding the timeFactor default but afaik that functionality has worked using the resolveTime method...

I turned debugging on this morning and notice this output where it aborts transition in my test app...

I pasted the output here: https://github.com/simplabs/ember-simple-auth/issues/574

jrjohnson:
OK - well resolving that conflict for me causes your changes in this PR to work perfectly.  Refresh is called and validated, token is restored.  Maybe issues are just in your app.

erichaus:
@jrjohnson what was the specific change you made?

Im starting to wonder if there is something wrong with my test app?

also what version of ember? im using 1.9.1

jrjohnson:
I pulled in your changes from this PR in addon/authenticators/jwt.js on a 1.13.2 app.

erichaus:
reopening with rebase

PR #70, opened 2015-06-23
adding support for extra params in the credentials object

cherta:
Hello José!

I'm developing an ember app for a .net backend and using your project for it. It happened to me that the .net infrastructure requires to send to the server token endpoint not only the identifier and the password but also some extra params. In my case the params are client_id and grant_type.

I did a minor change -I guess is minor- to `TokenAuthenticator#authenticate` and added the ability to retrieve extra params from the credentials object, I guess this is the easiest way to do it but I understand if you want to add an extra parameter and merge credentials with this new object.

I wait for your comments,
Gabriel

### DIFF
--- a/addon/authenticators/token.js
+++ b/addon/authenticators/token.js
@@ -149,7 +149,11 @@ export default Base.extend({
     var authentication = {};
     authentication[this.passwordField] = credentials.password;
     authentication[this.identificationField] = credentials.identification;
-    return authentication;
+
+    delete credentials.password;
+    delete credentials.identification;
+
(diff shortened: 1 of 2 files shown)


jpadilla:
@cherta Thanks for this PR but I don't think this is a use case I'd like to support. It'd be way easier for someone to implement a custom authenticator that handled the extra parameters.

cherta:
No problem @jpadilla, I figured out later that I could reopen the authenticator and override that specific behavior.

Thanks for the response

jpadilla:
@cherta great to hear, thanks!

PR #72, opened 2015-06-27
Added jwt-resolved authenticator which takes a token in lieu of credentials

### LINKED ISSUE #64: Silent login after user confirm
I've a service where user should confirm his account after signup.
Confirmation is made clicking on a link sent via email. Once the user clicks that link, I confirm his account server side and return a token to the client. 
How can I set that token as the authorize token and then silently login the user?
I've tried setting session variables but I wasn't able to 'tell' to simple-auth to use that value as the authorizer token to use in following requests.

jrjohnson:
Extends the JWT Token Authenticator to take an existing token instead of credentials.

```
let authenticator = 'simple-auth-authenticator:jwt-resolved';
this.get('session').authenticate(authenticator, token);
```

Fixes #64 for example.  I needed this functionality because I have a pluggable authentication system on my backend.  While it CAN take a username / password it has other methods which give me access to a token without needed to submit a form.

This could be done in a custom authenticator as well I suppose, but I figure it might be useful for others.

I wasn't able to figure out how to test if authentication had worked so I left the 'authenticate takes the token' test commented out in case you knew what things I could assert there.  Otherwise I can just pull it out and rebase this PR.

I couldn't come up with a better name for this.  Any thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,27 @@ token = {
 
 An automatic token refresh request would be sent out at token[Config.tokenExpireName] - now(). A good practice with regards to token refreshing is to also set a "leeway", usually no more than a few minutes, to account for clock skew when decoding JSON Web Tokens in the server-side. Some libraries like [PyJWT](https://github.com/jpadilla/pyjwt) and [ruby-jwt](https://github.com/progrium/ruby-jwt) already support this.
 
+**Resolved JWT Authenticator**
+
+Extends the JWT Token Authenticator to take an existing token instead of credentials
+
+```js
+// app/controllers/login.js
+import Ember from 'ember';
+
+export default Ember.Controller.extend({
+  actions: {
+    authenticate: function() {
+      //Assumes we have a token from somewhere else
+      let token = this.get('token');
+      let authenticator = 'simple-auth-authenticator:jwt-resolved';
+
+      this.get('session').authenticate(authenticator, token);
+    }
+  }
+});
+```
+
 ## The Authorizer
 
 The authorizer authorizes requests by adding `token` property from the session in the `Authorization` header:
--- a/addon/authenticators/jwt-resolved.js
+++ b/addon/authenticators/jwt-resolved.js
@@ -1,0 +1,42 @@
+import Ember from 'ember';
+import JwtTokenAuthenticator from './jwt';
+
+/**
+  Special type of JWT token that takes a token instead of credentials
+
+  The factory for this authenticator is registered as
+  'simple-auth-authenticator:jwt-resolved` in Ember's container.
+
+  @class JWTResolved
+  @namespace SimpleAuth.Authenticators
+  @module simple-auth-token/authenticators/jwt-resolved
+  @extends TokenAuthenticator
+*/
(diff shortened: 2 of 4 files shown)


jrjohnson:
Turns out my testing environment was a little wonky - for some reason this authenticator is calling login when it is restored.  I will track that down, but this isn't ready yet for sure.

jpadilla:
Although tests seem to be failing, this looks interesting.

Also, another use case for this might be to authenticate after signup without requiring a second request to actually authenticate with the API.

jrjohnson:
The test failures seemed to match what is currently [failing in master](https://travis-ci.org/jpadilla/ember-cli-simple-auth-token/jobs/67234553#L500).  Something going on with resolver / loader in ember 2.0 I'm guessing.  If this looks like something you want to merge I will clean up my commented out test and rebase and also look around for a solution to the loader issue in 2.0.

jpadilla:
@jrjohnson yeah don't know what's up with that. I'm still not 100% sure if we should include this in this package, but I'm definitely up for considering it. @erichonkanen have thoughts to share on this?

jrjohnson:
Ok - I rebased, but I wasn't able to solve the resolver issue easily other than to figure out it seems to be a new issue with 2.0 beta2.  Let me know if you would like other changes or if I should just do this as a custom authenticator on my own.

erichaus:
@jrjohnson nice work, there is an ongoing issue regarding restore/resolve if that's where you ran into trouble... so far I haven't been able to resolve it but hopefully soon

https://github.com/jpadilla/ember-cli-simple-auth-token/issues/71
https://github.com/jpadilla/ember-cli-simple-auth-token/issues/68

jrjohnson:
I was hitting an issue with ember-beta, but that was fixed (ignored) in [cae06f80d8260f6823f122a9a8c3c9d1f3b2cad1].  As far as I know this is now ready, unless changes are desired?

erichaus:
@jrjohnson out of curiosity, what are your thoughts on merging this with the JWT authenticator? could we support both jwt token or credentials (somehow check which one)? I don't know if that even makes sense or if a system would ever want both at the same time but just curious

jrjohnson:
If you mean doing this as a configuration variable instead of a completely separate authenticator I looked at that, but I didn't see a way to pass in a configuration var into the ember-cli-simpleauth session.authenticate() method.  

This seemed like the easiest way to be able to toggle the authenticator without adding a lot of developer overhead.  Other than that I don't have a preference though so if I missed a way to do this let me know.    I thought about parsing the incoming credentials object for either username/pass or a token, but I was worried that actually made it harder to use.  I'm ok with being wrong about that.

erichaus:
@jrjohnson yeah I was thinking about parsing the incoming creds object and checking if its a jwt token or not.. if it is, proceed forward as you made, otherwise assume its credentials

I don't know if anyone else has thoughts on this... @jpadilla ?

jrjohnson:
I'm going to put together a second PR which does what @erichonkanen suggested and detected the credentials passed.  I think thats a pretty clean way to handle this.

PR #73, opened 2015-07-01
Updated authenticate method to use tokenExpireName

erichaus:


erichaus:
rebased https://github.com/jpadilla/ember-cli-simple-auth-token/pull/69

engwan:
@jpadilla When can we have this merged? It's a very major bug: tokens not being refreshed after a page reload (`authenticator.restore`).

PR #77, opened 2015-07-07
Allow authentication by token

jrjohnson:
For use in cases where we already have access to the token and just
need to create an authenticated session.  Credentials can be passed
which contain a token key instead of a username and password.

Replaces #72

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,28 @@ export default Ember.Controller.extend({
 });
 ```
 
+***Authenticating with existing token***
+
+If you already have a JWT from a previous request or some other source you 
+can use that to authenticate directly.
+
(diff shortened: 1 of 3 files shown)


jrjohnson:
@erichonkanen this implements your suggestion from #72 Is there anything else I should do to get this merged?

jrjohnson:
@jpadilla if you're not into merging this, let me know and I will create a custom authenticator in my application instead.  Thanks!

jpadilla:
@jrjohnson thanks for the bump. Gonna go ahead and close this now. It's definitely a good reference for the future, but think we shouldn't be adding to this package. Thanks again!

PR #80, opened 2015-07-23
Add optional rootName for authorizer: 'simple-auth-authorizer:token'

pedroufo51warick:
Add optional rootName for credentials when sending the
authentication data to the server.


leandrocp:
:+1: It could be mentioned on README.md under https://github.com/jpadilla/ember-cli-simple-auth-token#available-customization-options

jpadilla:
Thanks, but at this point I'd prefer to not include any more settings. It's easy enough to just create a custom authenticator overriding `getAuthenticateData()`. Makes sense?

PR #81, opened 2015-07-24
return raw xhr object to allow more flexibility with error handling

jamesdixon:
In some cases, `responseJSON` or `responseText` may be `undefined`. For example, if the API is down, jQuery will return `status: 0`, but `responseJSON` will be `undefined`. Currently, this will be difficult to deal with. Returning the raw `xhr` object will allow for more flexibility when determining how to deal with specific types of errors.

### DIFF
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -181,7 +181,7 @@ export default TokenAuthenticator.extend({
         });
       }, function(xhr) {
         Ember.run(function() {
-          reject(xhr.responseJSON || xhr.responseText);
+          reject(xhr);
         });
       });
     });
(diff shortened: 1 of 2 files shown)


jpadilla:
@jamesdixon I see what you mean, but I'm not sure we want for this to be the default behavior, plus its not what's expected of any of the other simple-auth authenticators.

Closing for now, but might reconsider at a later point.

@marcoow any thoughts you could share on this?

jamesdixon:
@jpadilla understand your hesitance. That said, you bring up a good point about the behavior of other authenticators. Maybe we integrate a common response object that's returned rather than the raw `xhr` object or only `responseJSON`/`responseText`? I do think the additional information is warranted. Lemme know and thank you!

marcoow:
The error returned from the authenticator when session authentication fails is really just part of the authenticator's API - there's nothing like a common or expected behavior here. You also need to keep in mind that there doesn't have to be a response object at all even - an authenticator might just use the Facebook API or torii or whatever - in all of these cases there is no response object at all.

PR #97, opened 2015-10-19
Trigger sessionDataInvalidated on reject of a refresh

albertovasquez:
When rejecting we are not triggering the sessionDataInvalidated.  This event is handled by session and eventually redirects client to RootURL.


jpadilla:
@albertovasquez I think I ran into wanting this for the demo app.

leo:
@jpadilla Could you please merge this? Looks like a really awesome addition to me! :blush:

jpadilla:
Won't be able to merge this as is. We'll probably have a method where by default we'll only invalidate if status is 401 or 403. Introducing this as v2 since it's probably a breaking change. This method will be public and users the want the old behavior can just overwrite it.

PR #98, opened 2015-10-26
Add correct namespace for new version of ember-simple-auth.

paulchubatyy:
Recent version has probably changed that.

See this example: https://github.com/simplabs/ember-simple-auth#authorizers


MattNguyen:
I believe work is currently being done on this - https://github.com/jpadilla/ember-cli-simple-auth-token/pull/89.

PR #89, opened 2015-09-04
Catching up to ember-simple-auth jj-abrams branch

mmahalwy:
I'll fix the tests soon - just need to make time for it.

Issue #93, opened 2015-10-06
Ember 2.0 Support


aBuder:
Hi,

when we will have an Ember 2.0 implementation?

cresg820:
+1

ghost:
+1

dmitryt:
+1

erichaus:
anyone know what we need to do specifically to bring this into ember 2+? ive been away but using this lib for big project and may be able to help push it forward soon...

jpadilla:
@erichonkanen it'd be really helpful if you checked #100 out

jpadilla:
#100 was merged and is in master. Would love anyone interested to check it out.

PR #100, opened 2015-10-29
Continued work on updating to Ember-Simple-Auth 1.0

acorncom:
This continues the work on #89 slightly


listepo:
:+1:

wayne-o:
Just pulled this and ran ember test and all tests passed?

wayne-o:
OK npm test shows the failures - having a look now - not sure I totally understand what's going on though - will do my best

acorncom:
wat! ... ;-)  The token tests were all failing on my end when running `ember t -s`  Seems like the same issue came up on Travis too?

wayne-o:
Yeah - sorry - was running the wrong cmd :p it's complaining about not being able to load simple-auth-token/authenticators/token etc - any idea why?

acorncom:
I looked at it for a bit and thought "this could take a while".  The job I need this for doesn't have lots of extra cushion, so I decided to punt on it for a bit.  So no, hadn't tracked down why things were breaking

wayne-o:
if I change them to look for ember-cli-simple-auth-token it seems to hang right... is this a problem?

wayne-o:
scrap that - that also doesn't work :p

acorncom:
I was following this tutorial: http://log.simplabs.com/post/131698328145/updating-to-ember-simple-auth-10

wayne-o:
OK got the default tests passing - release, beta and canary still failing... I have 1:12 to fix it before I go home ;)

wayne-o:
These same tests fail on master...

acorncom:
@wayne-o Want me to add you to my Github repo so you can write to it directly as we work on that branch?

acorncom:
@wayne-o Updated: Just added you to my Github repo so you can write to it directly as we work on that branch

wayne-o:
I think we are going to need some help with this as I don't understand addons enough and the failing tests are a bit vague for me to get to the bottom of :/

RuslanZavacky:
@acorncom @wayne-o I am willing to help :) as we very need this addon to be working :)

acorncom:
@RuslanZavacky Just added you as a collaborator on my fork, feel free to send in commits.  My next gap to work on this is probably Thursday ...

acorncom:
@jpadilla If you've got time to review this, would love to get your input on the errors we're getting here ...

RuslanZavacky:
@acorncom sorted out, seems like good to go. I'll try to test in real life project and will see how it works.

RuslanZavacky:
It can't be merged for now, can't make it work in the real project..

wayne-o:
What's the status with this? It's beginning to be a blocker for me but I am not offay enough to help out all that much :/

RuslanZavacky:
@wayne-o actually you can, get this branch and test :) for me its working fine right now. Just spending more time validating everything else.

RuslanZavacky:
I've rebased it, so if tests will pass now, it means its good to go. But I suggest to wait 1-2 days, while I'll finish with authorization and then I can mark it as working )

jpadilla:
Thank you all for helping make this happen. I'll try to give this a go in the next couple of days. Currently working on other projects that aren't using this so I'm not as active, but I've been around and keeping an eye on all this activity.

mmahalwy:
is this good to go?

acorncom:
@mmahalwy Needs additional vetting, but it's close.  I'll be checking it on my end tomorrow.  If you find bugs, let me know and I can add you to my forked repo to make further changes :-)

RuslanZavacky:
@jpadilla I am a bit concerned about how we work with time.

We request from server, `expiresAt` - its a time, when our token intended to expire. And its not expressed as "how many seconds left", but instead its a date in future. So if server is in one Timezone, but client in browser in another timezone, it won't work. Because I can literally live "in future" comparing to my server. And then, token will refresh all the time.

Best thing to do, is return from server how much seconds left till token expires. And I see no use in in until this thing is solved, but it might involve more changes and will have breaking change.

Thoughts?

acorncom:
For the record, I've just upgraded our app to Ember-Simple-Auth 1.0.1 and this branch and all tests are passing.  We've got a fairly reliable acceptance test suite, so I do believe this is handling this well.  I'll make a few notes here in a minute documenting the changes I needed to make (and some of the ones I think we should make to this add-on)

RuslanZavacky:
@acorncom if you've upgraded to this commit "c4cd02d9e54208cba73d5d21e7fe8323a4a16617" I suppose you don't have tests related to token Refresh, because I did Breaking Change, as of this plugin do not compare server time and client time, but just uses seconds left till expiration.

Server API needs to be updated accordingly.

Tests won't show it, because from server you anyway return seconds, but now for example 300 seconds = after 5min refresh token, in old behaviour, you will return Unix timestamp, which will result in never ever refreshing token :)

acorncom:
@RuslanZavacky No, I don't have tests of the token refresh system yet, as our app is still in process of being launched and we're focusing on the frontend first :-)  Thanks for the heads up ...

acorncom:
**Background**
We store a few additional properties (in addition to `token`) on the result we return from the server to speed up lookups / setup.

Things I needed to change:
- changed `this.session.get(<property>)` to `this.get('session.data.authenticated.property')`.  Previously you had access to a raw JSON object directly on the session, but it now runs through a computed property which changes how we access things
- `config/environment.js#simple-auth` config key was changed to `ember-simple-auth`
- test helpers all need to be imported into each acceptance test using them and require the application to be passed in to each test helper

A few items we might want to change:
- old authenticators / authorizers were labeled `simple-auth-authenticator:jwt` or `simple-auth-authorizer:token`.  The new format is just to call it `authenticator:jwt` or `authorizer:token`
- might want to change our `config/environment.js` key to match by calling it `ember-simple-auth-token`
- might want to rename the npm package to match by calling it `ember-simple-auth-token`

Additional details on the upgrade can be found here (and most of the steps need to be followed): http://log.simplabs.com/post/131698328145/updating-to-ember-simple-auth-10

RuslanZavacky:
@acorncom unified naming across addon :) Renaming github package is up to @jpadilla. But yea, I agree that it should follow new standards, dropping -cli :)

listepo:
:+1:

RuslanZavacky:
I've found one bug with refresh token, if I am lucky, I'll fix it today, if not, then only on Monday :) stay tuned, but unless its fixed, this branch is't good to go :)

RuslanZavacky:
@acorncom @jpadilla guys, it seems like good to go. I've tested in my project. I've fixed a lot of stuff and did some breaking changes. Please review it and it possibly needs some changes in README.

acorncom:
@RuslanZavacky Other than the refresh changes, what other breaking changes have you made?  That seems like a good breaking change to me, but would be worth making sure we document any other ones ... (and have other people test your new work)

RuslanZavacky:
@acorncom timeFactor now means a bit different thing - it corrects expiration time returned from server. So if your server returned seconds, timeFactor will convert them to milliseconds, so they can be used in conjunction with javascript Date object.

jpadilla:
@RuslanZavacky I'm taking some time tonight or in the weekend to take this out for a spin. Great work for putting all this together.

coladarci:
@RuslanZavacky - I just saw your comment from above - I am strongly against this change as it goes against the JWT spec. If I'm looking at the wrong spec (https://tools.ietf.org/html/rfc7519#section-4.1.4), or I'm missing something, please let me know, but this if this update ships, I think all server-side jwt solutions will have to be hand-rolled. A serious deal-breaker for most users, I'd imagine.

jpadilla:
@RuslanZavacky I'm with @coladarci on this. Seems like diverting from the spec like this might be a bit too much, and would do better in a custom implementation for your use case.

RuslanZavacky:
I will revert exp field back, but will add ttl field, later on. 

But meanwhile, question - you are also so obsessed with spec, even though, this spec is so wrong.. am I the only one, who sees the problem with sending server time to client, and then compare it with a time of client and ignoring differences between client <-> server?

2468ben:
I also agree that however wrong/not wrong the spec is, the main bottleneck for folks right now is a tagged  version that's ember-simple-auth 1.0 compatible. So a PR with the breaking changes would let the compatible version come out, then another version with breaking changes could come when everyone agrees on those.

2468ben:
Also are the client & server not already communicating time as UTC? I thought that's how it worked to avoid the timezone issue.

RuslanZavacky:
@2468ben basically, you are right. In spec it says that time should be in UTC, I've overlooked it. Even though, all this time conversions complicates all implementation, instead of just sending how much seconds left :) If noone objects, I'll just additional field to exp -> ttl. And this will satisfy both needs.

RuslanZavacky:
looking even more into it, nothing needs to be changed about exp, I'll revert it soon :)

addon/authenticators/jwt.js
```diff
@@ -168,7 +169,7 @@ export default TokenAuthenticator.extend({
         Ember.run(function() {
           var token = response[_this.tokenPropertyName],
             tokenData = _this.getTokenData(token),
-            expiresAt = tokenData[_this.tokenExpireName],
+            expiresAt = _this.getCurrentTime() + tokenData[_this.tokenExpireName] * _this.timeFactor,
```
coladarci: this is still wrong, right @RuslanZavacky ? As discussed, the token will contain the expiration time, not how long from now it expires.

addon/authenticators/jwt.js
```diff
@@ -240,10 +237,10 @@ export default TokenAuthenticator.extend({
     return new Ember.RSVP.Promise(function(resolve, reject) {
       _this.makeRequest(_this.serverTokenRefreshEndpoint, data).then(function(response) {
         Ember.run(function() {
-          var token = response[_this.tokenPropertyName],
-            tokenData = _this.getTokenData(token),
-            expiresAt = tokenData[_this.tokenExpireName],
-            tokenExpireData = {};
+          var token = response[_this.tokenPropertyName];
+          var tokenData = _this.getTokenData(token);
+          var expiresAt = _this.getCurrentTime() + tokenData[_this.tokenExpireName] * _this.timeFactor;
```
coladarci: this is still wrong, right @RuslanZavacky ? As discussed, the token will contain the expiration time, not how long from now it expires.

RuslanZavacky: I am still in progress. basically almost reverted whole this file.

cliedelt:
Any Update?

RuslanZavacky:
@jpadilla referring to #93 - I am using this library with Ember 2.0, so it works :) how about moving forward with it and merging?

cliedelt:
@RuslanZavacky :+1:

mvrkljan:
@jpadilla any info on when we might expect this to be merged and available via cli?

addon/authenticators/jwt.js
```diff
@@ -103,7 +104,7 @@ export default TokenAuthenticator.extend({
       dataObject = Ember.Object.create(data);
 
     return new Ember.RSVP.Promise(function(resolve, reject) {
-      var now = (new Date()).getTime();
+      var now = _this.getCurrentTime();
```
jpadilla, line 107: @RuslanZavacky Why is `getCurrentTime()` here useful? Why would you override it?

addon/authenticators/token.js
```diff
@@ -146,10 +146,16 @@ export default Base.extend({
     @return {object} An object with properties for authentication.
   */
   getAuthenticateData: function(credentials) {
-    var authentication = {};
-    authentication[this.passwordField] = credentials.password;
-    authentication[this.identificationField] = credentials.identification;
-    return authentication;
+    var password = credentials.password;
```
jpadilla: @RuslanZavacky Any specific reason for rewriting this block? Not a big fan of modifying the `credentials` object that's passed in by doing `delete ...`.

jpadilla:
@RuslanZavacky left some inline comments. Also noticed something funky running the demo app.
1. Login
2. Wait for token refresh to start running
3. Go to another tab for x seconds
4. Endpoint returned a 500 error because token expired. This doesn't seem to happen if I stay on the same tab.

Thoughts on what might be going on? If I recall something like this happened at some point before. @erichonkanen do you recall what happened then?

addon/authenticators/jwt.js
RuslanZavacky, line 107: Abstraction for tests, we need to have defined state. Time is always random state in tests, which is not good.

addon/authenticators/token.js
RuslanZavacky: good question, have no idea what I was doing here. Maybe too much wine :)

acorncom:
Are we ready to merge this at this point?  I can help with further tidy up if needed in the next week or so ...

jpadilla:
@acorncom talked to @RuslanZavacky about checking out https://github.com/jpadilla/ember-cli-simple-auth-token/pull/100#issuecomment-160284249. That's all that's left that I can see.

RuslanZavacky:
I'll check it on weekend. Currently attending conference :)

RuslanZavacky:
@jpadilla I've fixed Demo app. Seems like it did not work before either?

Also I've added ability to pass additional headers for authentication request, useful, when you attach social account for already authenticated user :)

jpadilla:
@RuslanZavacky awesome, although I'm still seeing the issue with the expiring tokens. Managed to [record](https://monosnap.com/file/L40k1OgeV36t4JzYcrndfQut1vINzE) it.

RuslanZavacky:
@jpadilla hm, it might make sense. When you switch tab, Chrome stops execution, so refresh token can't be obtained. Token life is 10 seconds, refresh leeway is 5, so if you did not switch back in ~ 5 seconds, token expires. And if token is expired, you can't refresh it, you need to obtain the new one.

jpadilla:
@RuslanZavacky yeah that's what I thought. Could we handle the expiration gracefully on the dummy endpoint and have it return a valid response?

jrjohnson:
I think this has been an existing issue. We have a really big refresh leeway in our app to help with this.

RuslanZavacky:
@jpadilla I'll take a look :)

jpadilla:
@jrjohnson thanks for confirming

RuslanZavacky:
@jpadilla I think it should be a separate ticket, seems like there is no 'simple fix'

addon/authorizers/token.js
```diff
@@ -73,27 +75,14 @@ export default Base.extend({
...
+  authorize: function(data = {}, block = () => {}) {
+    const token = data[this.tokenPropertyName];
 
     if (this.get('session.isAuthenticated') && !Ember.isEmpty(token)) {
-      if(this.authorizationPrefix) {
-        token = this.authorizationPrefix + token;
-      }
-
-      jqXHR.setRequestHeader(this.authorizationHeaderName, token);
+      block(this.authorizationHeaderName, this.authorizationPrefix + token);
```
andruby: Is it possible that this causes a regression when `authorizationPrefix` is set to `null`?

Using this branch and `authorizationPrefix: null` the authorization headers generated are `nullYekatoken..`.

RuslanZavacky: yeap, you are right. Need to fix that :)

andruby: Thank you for your quick & swift solution!

listepo:
Ping

jpadilla:
I'll be working on updating ember-cli and a few things before releasing a new version. Thanks everyone and specially @RuslanZavacky for all your awesome work.

RuslanZavacky:
:+1:

mmahalwy:
woot!
@coladarci

richardvenneman:
:pray:

jpadilla:
@RuslanZavacky @acorncom are there any documentation updates we need to do?

jpadilla:
Just published v1.0.0 https://www.npmjs.com/package/ember-simple-auth-token

PR #112, opened 2015-12-23
Use Ember.get to fetch properties in JWT response

xcambar:
This allows for more flexibility in the JWT structure.


jpadilla:
@xcambar thanks!

PR #114, opened 2015-12-27
Update README.md

rmzr7:
updated package name


jpadilla:
Package will be published under `ember-simple-auth-token`.

PR #122, opened 2016-01-25
Add warning when refreshLeeway is set to high

albertovasquez:
When the expiresAt is in the future we still might skip it if the refreshLeeway is set to high.
In lieu of running the refreshAccessToken method and possibly cause non stop calls to refresh endpoint
lets add a warning to let admin know why their tokens are not refreshing.

### DIFF
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -207,6 +207,10 @@ export default TokenAuthenticator.extend({
         delete this._refreshTokenTimeout;
 
         this._refreshTokenTimeout = Ember.run.later(this, this.refreshAccessToken, token, wait);
+      } else if (!Ember.isEmpty(token) && !Ember.isEmpty(expiresAt) && (expiresAt - now > 0)) {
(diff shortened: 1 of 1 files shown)


eluciano11:
Hi @albertovasquez , thank you for your PR. I got around to looking at this today. I don't really understand how `(expiresAt - now > 0)` indicates that the refreshLeeway is too large. I was thinking about this problem and shouldn't it be `((this.refreshLeeway * 1000) > (expiresAt - now))` ? The default `refreshLeeway` is 0 but it can be configured by the user. So if the `refreshLeeway` is greater than the time we have can subtract from the token expiration time than it's too big.

Please let me know what you think and the thought process behind `(expiresAt - now > 0)` so we can add this warning as soon as posible.

Thank you for your work.

fenichelar:
Great idea! Due to the depreciation of `Ember.Logger`, now if `refreshLeeway` is too large, an error is thrown: https://github.com/jpadilla/ember-simple-auth-token/blob/master/addon/authenticators/jwt.js#L171-L172

PR #124, opened 2016-01-31
Upgrade CLI & ember. Support for nested tokenPropertyName

thhermansen:
This pull request main goal was to add support for nested `tokenPropertyName`. Instead of having `token` as value we use `some.nested.token`. The `authenticate()` did already support this, as it used `Ember.get(obj, tokenPropertyName)`, but other parts of the code didn't support this.

I also added a fix for issue I encountered in `getTokenData()` when decode the JWT token when it contained UTF-8 encoded characters. 

A bit sorry for updating Ember CLI in this pull request. I didn't manage to get `ember test` to run successfully with a newer version of Node without do this upgrade. Here are some links which may be relevant to this issue:
- https://github.com/ember-cli/ember-cli/issues/5110
- https://github.com/ember-cli/ember-cli/issues/5317


eluciano11:
Hi @thhermansen , thank you for your work! I've made another PR on #145 with your commits. I removed the ember and ember-cli upgrades because we had merged and update already. 

The test are not freezing anymore but I did find a strange behaviour on the tests added. For some reason they fail on development but not with Travis CI. If you run `ember test` they fail but if you run `ember test --filter='#refreshAccessToken triggers the`sessionDataUpdated`event on successful request.'` they work. 
Any ideas why this is happening? 

Feel free to add any changes on #145 .
Again, thank you for your time and work.

thhermansen:
Please see #145 on updates why some test failed.

PR #127, opened 2016-02-10
Update JWT refreshAccessToken to behave more like authenticate.

jribar:
I'm overriding the handy getResponseData function to stash some additional attributes returned in the token, in the session.  This works great.  However when the token is refreshed, getResponseData is never called.

This patch updates refreshAccessToken to behave like the initial authenticate function.


eluciano11:
Hi @jribar , thank you for your PR. I had the opportunity to go over it today. I believe that the behaviour you implemented was added indirectly in #145. The PR was merged yesterday. We plan to release a new version of the add-on later today or tomorrow.

Thank you for your work!

PR #134, opened 2016-03-13
Add bug fix for JWT restore

digia:
JWT restore's refreshAccessToken should `resolve` with the data it's
provided, not with the original data. Restoring with the original data doesn't
update the token in the store.


eluciano11:
Hi @digia , thanks for the PR. We will merge your changes in #140. We had to upgrade the ember-cli version. For some reason the tests were not working properly.

PR #141, opened 2016-03-25
Add fix for expiresAt

digia:
exp should be in seconds not milliseconds. According to the spec - https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-32#page-6

### DIFF
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -311,6 +311,11 @@ export default TokenAuthenticator.extend({
     return new Ember.RSVP.resolve();
   },
 
+  /**
+    Returns the current time as a timestamp in seconds
+    @method getCurrentTime
+    @return {Integer} timestamp
+  */
   getCurrentTime() {
     return Math.floor((new Date()).getTime() / 1000);
   },
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -312,7 +312,7 @@ export default TokenAuthenticator.extend({
   },
 
   getCurrentTime() {
-    return (new Date()).getTime();
+    return Math.floor((new Date()).getTime() / 1000);
   },
 
   /**


addon/authenticators/jwt.js
```diff
@@ -312,7 +312,7 @@ export default TokenAuthenticator.extend({
   },
 
   getCurrentTime() {
```
jpadilla, line 319: @digia mind adding a doc block for this method?

jpadilla:
**NOTE: THIS IS A BREAKING CHANGE**

digia:
@jpadilla Doc blocks are live!

eluciano11:
@digia awesome! We'll ship this on the 2.0 release. We're getting around all the bug fixes that we can to make release a 1.1.0 before.

Subtletree:
With this fix should the `wait` calculation in `scheduleAccessTokenRefresh` and `restore` be changed also?

At the moment the `wait` time is calculated like so:

 `const wait = expiresAt - now - (this.refreshLeeway * 1000);`

It seems the `refreshLeeway` is being converted to milliseconds and this should be left as seconds.

Since `Ember.run.later` expects the interval to be in milliseconds the whole equation should be multiplied by 1000, not just the leeway e.g:

`const wait = (expiresAt - now - this.refreshLeeway) * 1000;`

joshsmith:
Any possibility for a 2.0 branch or similar that this can be looped into?

Subtletree:
This would be great.

In the meantime I'm just overriding the relevant methods. Here's my [custom jwt authenticator](https://gist.github.com/Subtletree/e3ae75f0d378732d4aa626d0aeea6da9) if it's any use

The methods are just copied and have only the timing error changed

joshsmith:
Thanks for sharing @Subtletree!

mansona:
Hey Folks, 

I just came across this issue running against a server-side JWT implementation that sets exp to seconds (like the spec). Is there any movement on getting this into a branch ahead of the v2.0.0 release? 

Also does the milestone cover everything that is needed to get a release out? https://github.com/jpadilla/ember-simple-auth-token/milestone/1 just wondering so that I can help if possible 👍

Ramblurr:
I'm in the same boat as @mansona on this one. Just ran into this issue and am quite sad it's not yet released.

It's been almost 10 months since this issue was reported and a PR submitted.

This is obviously a backward incompatible change. I would be perfectly happy if 2.0.0 was released with just this fix.

I'd be more than happy to help move this along. What else is blocking to get this out the door?

jpadilla:
Let's do this. Sorry for the hold up, and thanks to @digia and everyone else! 🎁

jpadilla:
Just published v2.0.0 with this change!

jpadilla:
Might be a good idea to leave `timeFactor` behind https://github.com/jpadilla/ember-simple-auth-token/search?utf8=%E2%9C%93&q=timeFactor

If anyone wants help out, that'd be awesome!

Subtletree:
Thanks @jpadilla !

Think there's a couple lines that need to be changed for calculating the wait time before tokens are refreshed. Just submitted pr #181

rrglomsk:
It looks like the times calculated for `expiresAt` and `now` are in two different units of time now, since `resolvesTime` returns milliseconds and `getCurrentTime` now returns seconds... 😳

Subtletree:
Oh yeah, looks like `resolvesTime` returns seconds as long as `timefactor = 1` and `exp` is in seconds, but not if either of those are using milliseconds.

Getting rid of `timeFactor` sounds like a good solution, and just sticking to the jwt spec having everything in seconds. Maybe I'll take a look at that.

rrglomsk:
`return new Date(time * this.timeFactor).getTime();`

`resolvesTime` will always return milliseconds because [getTime](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime) returns milliseconds heheh

Subtletree:
Check this fiddle https://jsfiddle.net/gd4sg6z2/.

`getTime` does return milliseconds but if you pass a number of _seconds_ to `Date`, `Date` assumes they are milliseconds so `getTime` just returns the same value. Seems unnecessary to have `Date` in there at all? Could have just had `return time * this.timeFactor`?

Anyway i'm about to push a pr that removes `resolvesTime` altogether so shouldn't be an issue 😄

rrglomsk:
oh yeah!! Thanks for pointing that out 👍  And sounds good 😄

Subtletree:
OK made a new pull request #182. Haven't made a whole lot before so let me know if I've missed something!

Issue #135, opened 2016-03-22
Authenticate with form parameters


jgadbois:
Our server side authentication expects for parameters for the login instead of JSON.  After authentication everything uses JWT.

Is there anyway to pass username and password initially as form parameters?

eluciano11:
Hi @jgadbois , I think this issue can be solved by extending the authenticator that the add-on provides. Currently we provide `jwt` and `token`. If you're using jwt based authentication you need to import and extend the authenticator. After you extend it you can modify the `makeRequest` function to send the data in form parameters instead of json.

Example:

Create a folder named `authenticators` in you `app`. In it create a js file example: `custom.js`. The file name will be used when you use the `authenticate` function.

``` javascript
//app/authenticators/custom.js

//If you're using jwt you should use the jwt authenticator.
import JWTAuthenticator from 'ember-simple-auth-token/authenticators/jwt';

//If you're not using jwt you should use the token authenticator
//import TokenAuthenticator from 'ember-simple-auth-token/authenticators/token';

export default JWTAuthenticator.extend({
  makeRequest: function(url, data, headers) {
    return Ember.$.ajax({
      url: url,
      data: data,
      type: 'POST',
      beforeSend: (xhr, settings) => {
        xhr.setRequestHeader('Accept', settings.accepts.json);

        if (headers) {
          Object.keys(headers).forEach(key => {
            xhr.setRequestHeader(key, headers[key]);
          });
        }
      }
    });
  }
});

//app/login/route/route.js
import Ember from 'ember';

export default Ember.Route.extend({
  actions: {
    login(data) {
      const controller = this.controllerFor('login');

      controller.set('isLoginIn', true);
      // this.get('session').authenticate('authenticator:nameOfTheCustomAuthenticator', data);
      this.get('session').authenticate('authenticator:custom', data);
    }
  }
});
```

Hope this helps!

jgadbois:
Thanks for following up on this!

On Fri, Mar 25, 2016 at 8:45 AM -0700, "José Padilla" notifications@github.com wrote:

Closed #135.

PR #144, opened 2016-03-26
Allow generic options to the token authenticators

musaffa:
Consider this use case:

``` Javascript
signInWithGoogle() {
  const self = this;

  this.get('torii').open('google-oauth2').then(function (auth_code) {
    return self.get('session').authenticate('authenticator:jwt', auth_code);
  }).catch(function(error) {
   // error code goes here
  });
}
```

When a user grants access to his google account, Google oauth2 service sends back an authorization code. `auth_code` is a hash which contains three fields: `authorizationCode`, `provider`, `redirectUri`. The authenticator sends this hash at the token endpoint with which the server request the provider for the protected resources, creates a user account if needed, generates its own token and sends it back to the client.

Current `ember-simple-auth-token` restricts the hash to `password` and `username` only. That's why the above use case isn't possible. Removing the restriction will allow this addon to be used in use cases beyond just `password` and `username` scenario. It wont affect username/password authentication either. What is sent to the token endpoint will depend on what credentials is passed to the authenticate method.  

This PR removes the restriction and cleans up a lot of code related to the restriction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,8 +178,6 @@ For the Token authenticator:
 // config/environment.js
 ENV['ember-simple-auth-token'] = {
   serverTokenEndpoint: '/api/token-auth/',
-  identificationField: 'username',
-  passwordField: 'password',
   tokenPropertyName: 'token',
   authorizationPrefix: 'Bearer ',
   authorizationHeaderName: 'Authorization',
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -71,8 +71,6 @@ export default TokenAuthenticator.extend({
   init() {
     this.serverTokenEndpoint = Configuration.serverTokenEndpoint;
     this.serverTokenRefreshEndpoint = Configuration.serverTokenRefreshEndpoint;
-    this.identificationField = Configuration.identificationField;
-    this.passwordField = Configuration.passwordField;
     this.tokenPropertyName = Configuration.tokenPropertyName;
     this.refreshAccessTokens = Configuration.refreshAccessTokens;
     this.refreshLeeway = Configuration.refreshLeeway;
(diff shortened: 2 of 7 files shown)


jpadilla:
This looks like a breaking change and from the looks of it you might be able to get around to doing what you need by using `.getAuthenticateData()`.

musaffa:
A deprecation strategy can be adopted for the breaking change. In the intermediate step towards the migration, the addon should work with or without the restriction. I can lay my hand on it if you want.

jpadilla:
@musaffa I'm not currently convinced that we'd want or need this though. Like I said above, I think it should be possible to just use `.getAuthenticateData()`.

musaffa:
@jpadilla How can I use `.getAuthenticateData()` when it is there the restriction is implemented? The primary object of this PR is to remove that method!

eluciano11:
Hi @musaffa, `getAuthenticateData` is currently a public method. You'd need to extend the token authenticator and write your own `getAuthenticateData` function. We have an example on #135

musaffa:
Oh thanks. I'll try this out.

I have a question though. Why do u want to put restriction on the token attributes by default? Token credentials are much more than username and password, right?

eluciano11:
Well in one hand, I didn't create this add-on so I don't really know the thought process behind the default attributes. On the other hand, I've used the add-on a lot and in every case all I've needed is email and password for the authentication. From what I've seen a lot of people need the same fields as well. I think it's important to highlight the difference behind the token management. On your comment you suggested an OAuth2 use case. OAuth2 is different from JWT or another tokens. I don't think that this add-on was made to cover the OAuth2 use case and that's why it has identification and password fields by default. Also, ember simple auth provides [provides](https://ember-simple-auth.com/api/classes/OAuth2PasswordGrantAuthenticator.html) OAuth2 management out of the box. I've never used it but maybe that could cover your use case.

I will close this PR for now. Thank you for your work.

musaffa:
@eluciano11 The problem is the credentials that are allowed to pass at the token endpoint. Be it username/password or ouath2 or openid connect, it is the responsibility of the server to use those credentials to respond with token/jwt. This addon has the ability to cover all those use cases but, unfortunately, restricts the credential fields by default.

PR #145, opened 2016-03-27
Support for nested tokenPropertyName (modified)

eluciano11:
This pull request contains the commits made by @thhermansen on #124 . It does not include the upgrade to ember and to ember-cli. Upgrade to ember and ember-cli 1.13.15 were made on #136.

### DIFF
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -113,7 +113,7 @@ export default TokenAuthenticator.extend({
       if (Ember.isEmpty(expiresAt)) {
         // Fetch the expire time from the token data since `expiresAt`
         // wasn't included in the data object that was passed in.
-        const tokenData = this.getTokenData(data[this.tokenPropertyName]);
+        const tokenData = this.getTokenData(Ember.get(token, this.tokenPropertyName));
 
         expiresAt = this.resolveTime(tokenData[this.tokenExpireName]);
         if (Ember.isEmpty(expiresAt)) {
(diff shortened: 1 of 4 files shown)


addon/authenticators/jwt.js
```diff
@@ -158,24 +158,13 @@ export default TokenAuthenticator.extend({
                                  otherwise
   */
   authenticate(credentials, headers) {
-
     return new Ember.RSVP.Promise((resolve, reject) => {
-      const data = this.getAuthenticateData(credentials);
+      let data = this.getAuthenticateData(credentials);
```
jpadilla: Any reason why this can't be `const`?

addon/authenticators/jwt.js
```diff
@@ -226,26 +215,14 @@ export default TokenAuthenticator.extend({
     @private
   */
   refreshAccessToken(token, headers) {
-    let data = {
-      [this.tokenPropertyName]: token
-    };
+    let data = this.makeRefreshData(token);
```
jpadilla: Any reason why this can't be `const`?

thhermansen:
I think I fixed the failing test and changed `let` to `const`. Please see [this diff](https://github.com/eluciano11/ember-simple-auth-token/compare/master...thhermansen:eluciano11-master) between @eluciano11 master against my latest changes.

As far as I understood it, Qunit tear down and moved to the next test a little too fast and didn't wait for the test to complete it's assertions. I fixed this by doing an `assert.async();` and calling the async callack when the test was done.

Feel free to merge my changes is to @eluciano11 master branch so the changes will be included in to this pull request (as I don't think I can do that myself).

eluciano11:
@homu r+

eluciano11:
@thhermansen thank you for your work! Soon we'll release a new version of the add-on with your changes.

PR #146, opened 2016-03-27
Modifies and upgrades ember-try configuration

eluciano11:
Upgrades ember-try version and modifies its configuration to support user running ember versions from 1.11 to 2.0. It also modified the travis ci script that was being used.


eluciano11:
@homu r+

PR #147, opened 2016-03-28
Adds token validation on handleAuthResponse

eluciano11:
Adds token validation on handleAuthResponse. Users should not get an error Uncaught TypeError: Cannot read property 'split' of undefined

### DIFF
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -164,8 +164,13 @@ export default TokenAuthenticator.extend({
 
       this.makeRequest(this.serverTokenEndpoint, data, headers).then(response => {
         Ember.run(() => {
-          const sessionData = this.handleAuthResponse(response, reject);
-          resolve(sessionData);
+          const sessionData = this.handleAuthResponse(response);
+
+          try {
+            resolve(sessionData);
+          } catch(error) {
+            reject(new Error(error));
+          }
         });
       }, xhr => {
         Ember.run(() => { reject(xhr.responseJSON || xhr.responseText); });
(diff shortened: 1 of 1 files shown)


jpadilla:
I was thinking we could do something like this. We'd warn about empty token, and prevent the uncaught typerror. Thoughts?

``` javascript
getTokenData(token) {
  if (Ember.isEmpty(token)) {
    Ember.Logger.warn('TOKEN IS EMPTY OR A BETTER WARNING');

    return {};
  }

  const payload = token.split('.')[1];
  const tokenData = decodeURIComponent(window.escape(atob(payload)));

  try {
    return JSON.parse(tokenData);
  } catch (e) {
    return tokenData;
  }
},
```

eluciano11:
Yeah, I thought about managing it in the `getTokenData` function but the `restore` function was using something [similar](https://github.com/jpadilla/ember-simple-auth-token/pull/147/files#diff-b3b19e97b14851cfabb1e6558abfc6daL110) to what I implemented on `handleAuthResponse`. We could remove that validation from `restore` and manage it in `getTokenData` as suggested above.

addon/authenticators/jwt.js
```diff
@@ -220,9 +226,14 @@ export default TokenAuthenticator.extend({
...
-          const sessionData = this.handleAuthResponse(response);
-          this.trigger('sessionDataUpdated', sessionData);
-          resolve(sessionData);
+          try {
+            const sessionData = this.handleAuthResponse(response);
+
+            this.trigger('sessionDataUpdated', sessionData);
+            resolve(sessionData);
+          } catch(error) {
+            reject(new Error(error));
```
jpadilla: Should be `reject(error);`

addon/authenticators/jwt.js
```diff
@@ -163,8 +164,13 @@ export default TokenAuthenticator.extend({
...
       this.makeRequest(this.serverTokenEndpoint, data, headers).then(response => {
         Ember.run(() => {
-          const sessionData = this.handleAuthResponse(response);
-          resolve(sessionData);
+          try {
+            const sessionData = this.handleAuthResponse(response);
+
+            resolve(sessionData);
+          } catch(error) {
+            reject(new Error(error));
```
jpadilla: Should be `reject(error);`

eluciano11:
@homu r+

PR #148, opened 2016-03-29
fix deprecation Ember.merge for 2.5.0

gertjanwytynck:
DEPRECATION: Usage of `Ember.merge` is deprecated, use `Ember.assign` instead. [deprecation id: ember-metal.merge]

More info: emberjs/ember.js#12303

Setup:

```
Ember             : 2.5.0-beta.3
Ember Data        : 2.5.0-beta.3
jQuery            : 2.2.1
Ember Simple Auth : 1.1.0-beta.3
```

### DIFF
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -2,6 +2,8 @@ import Ember from 'ember';
 import Configuration from '../configuration';
 import TokenAuthenticator from './token';
 
+const assign = Ember.assign || Ember.merge;
+
 /**
   JWT (JSON Web Token) Authenticator that supports automatic token refresh.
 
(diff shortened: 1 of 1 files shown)


jpadilla:
I don't think our tests cover 2.5.0 beta or at least I can't see the deprecation anywhere on Travis, but this seems like a small enough change to accept. @eluciano11 thoughts?

eluciano11:
No, currently we don't cover the `2.5.0` scenario in our ember try config. We currently test against every version from `1.11` through `2.0.0`. We can change that to make tests against ember `1.13` through `2.5`. I can make those changes and add them to this PR.

eluciano11:
@homu r+

eluciano11:
@gertjanwytynck Thank you for your contribution!!

PR #150, opened 2016-04-03
Adds token invalidation when the server responds with 401 or 403

eluciano11:


server/mocks/api/token-refresh.js
```diff
@@ -8,7 +8,7 @@ module.exports = function(app) {
     jwt.verify(req.body.token, 'secret', function(err, decoded) {
       if (err) {
         res
-          .status(401)
+          .status(402)
```
jpadilla: why change this?

eluciano11: mmm typo on testing I guess...

addon/authenticators/jwt.js
```diff
@@ -238,6 +238,7 @@ export default TokenAuthenticator.extend({
           }
         });
       }, (xhr, status, error) => {
+        this.handleTokenRefreshFail(xhr.status);
         Ember.Logger.warn(`Access token could not be refreshed - server responded with ${error}.`);
```
jpadilla, line 242: Small nitpick, let's log first and then do `handleTokenRefreshFail`

basz:
Q: Why only for 401 and/or 403? According to the [specs](https://tools.ietf.org/html/rfc6749#section-5.2) 400 might be returned too when the refresh token has expired ('invalid_grant')? 

ps. This repo is NOT about the default oauth2 implementation in ember-simple-auth is it?

jpadilla:
@basz right, this isn't about the oauth2 implementation in ESA. For now I'd rather have this on 401/403. @eluciano11 thoughts?

jpadilla:
@eluciano11 we also need some docs on this.

eluciano11:
Yeah, I decided to use 401 and 403 because this will be called on the token refresh. When a server returns a `Forbidden (403)` it mean it understood the request but it doesn't authorise it. When the client sends any other request with an unauthorized token the server will respond with `Unauthorized (401)`.

HTTP 400 is a `Bad Request` and I don't know if that is very common in this scenario. I've used Django Rest Framework with JWT and from my experience the backend usually returns a `403` when the token isn't refreshed.

If your backend returns a HTTP 400 when the token fails to refresh you can extend the JWT authenticator and redefine the `handleTokenRefreshFail` function.

basz:
400 is in the oauth2 spec. 

Extending is acceptable though

Thx

PR #152, opened 2016-04-04
Invalidate on idle functionality

eluciano11:
This PR contains the functionality to invalidate a token after a user has been idle for a specified period of time. The implementation depends on the `ember-user-activity` add-on. Users of the add-on can specify the amount of time that has to pass for a user to be considered idle (5 minutes by default) on the property `invalidateAfter`. The implementation observes the `isIdle` property provided by the `userIdle` service. If `isIdle` changes to `true` and the user of the add-on has set the `invalidateIfIdle` to `true` then the `invalidate` function will be called to invalidate the token and the `sessionDataInvalidated` will be triggered.


jpadilla:
@eluciano11 should we pick this up sometime this week?

eluciano11:
yes!

jpadilla:
@eluciano11 this also needs some docs.

PR #153, opened 2016-04-12
Fixed a bug with restoring token data from local storage

acorncom:
As part of 1.1.0, a new bug was introduced with restoring token data from local storage (the line in question was changed in https://github.com/jpadilla/ember-simple-auth-token/pull/145/commits/ba23ed4741f5d7048478555b113ba58da545f0f2 )

This tweaks what object the token is being retrieved from to use the pre-fetched token (instead of trying to fetch the token from the token 😄 )

### DIFF
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -116,7 +116,7 @@ export default TokenAuthenticator.extend({
       if (Ember.isEmpty(expiresAt)) {
         // Fetch the expire time from the token data since `expiresAt`
         // wasn't included in the data object that was passed in.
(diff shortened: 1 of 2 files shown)


jpadilla:
@acorncom ace! thank you very much.

jpadilla:
@homu +r

jpadilla:
@homu r+

jpadilla:
@acorncom just published `v1.1.1`, thanks again!

acorncom:
@jpadilla Thank you!  I hit the bug on our end and it took a bit to track down.  Showed up here to find a patch already, only to find it still had new wrinkles post-updated ;-) That's why I sent in the test too ;-)

Issue #104, opened 2015-11-08
Invalidating the session on failed refresh


angelosarto:
When the token refresh endpoint fails to refresh the token and returns a status such as 403, 500, etc.  The token appears to remain in local storage.

Is this the expected behavior?
Is there a place where I can hook this to invalidate the session if the refresh fails?  

This also seems to be the case if it just times out rather than a failure, so if I get a token but don't refresh it the session sticks around.

albertovasquez:
@angelosarto take a look at this PR - https://github.com/jpadilla/ember-cli-simple-auth-token/pull/97 I had the same question and resorted to this for now.

jpadilla:
I ran into wanting this for the demo app. Honestly the only reason why we're not currently doing that is because simple-auth's oauth authenticator doesn't do that. But now I'm wondering why it doesn't.

@marcoow is there any reason why the oauth authenticator doesn't trigger `sessionDataInvalidated` for example whenever `_refreshAccessToken` is rejected?

marcoow:
The thinking behind not triggering `sessionDataInvalidated` when refreshing the access token fails is that the fact that refreshing fails doesn't mean the access token that the app already has is invalid. One could argue that certain responses indicate that the session should be invalidated of course (e.g. 401 and 403). I'd say though that a 500 error should not lead to `sessionDataInvalidated` being triggered.

jpadilla:
@leo we were talking about this a few weeks ago, did we arrive to any conclusions?

leo:
@jpadilla Yes. It is definitely a great feature to be implemented. However, it's important to make sure that the session only gets invalidated if the server returns a 401 or 403 error (not for 500, for example - like Marco said it).

huntie:
Hey all. Would really like to see this functionality in the core library soon, and so have submitted the pull request above for review.

mblayman:
Is there anything left for this issue? It looks like the related bits have been merged. Am I missing anything?

mblayman:
Looks like #150 invalidates the session on 401 and 403. There was some discussion about invalidating on 400 and it was left as an exercise to the implementer to customize `handleTokenRefreshFail` to handle 400 if desired.

As a datapoint for consideration, DRF JWT will return 400 if the "Signature has expired". I don't know if one common backend is enough to justify adding 400 into the mix.

fenichelar:
[...]

PR #160, opened 2016-05-22
Trigger sessionDataInvalidated event on failed token refresh

huntie:
Invalidate the session if server responds with client errors 401 Unauthorized or 403 Forbidden upon refreshing an expiring JWT. As outlined in #104.

### DIFF
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -239,6 +239,11 @@ export default TokenAuthenticator.extend({
         });
       }, (xhr, status, error) => {
         Ember.Logger.warn(`Access token could not be refreshed - server responded with ${error}.`);
+
+        if (xhr.status === 401 || xhr.status === 403) {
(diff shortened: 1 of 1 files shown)


kris-ellery:
Any idea when this pull request will be merged and released?

jpadilla:
@KrisOlszewski thanks for the follow up.

kris-ellery:
@jpadilla no problem. I wasn't sure which implementation you were going to use, this one or #150 Both passed checks and were ready to merge.

jpadilla:
@KrisOlszewski grrr good catch, I think I meant for #150 to be the one since it has tests.

kris-ellery:
@jpadilla makes sense, seems like a more robust solution. I look forward to the next release!

jpadilla:
@KrisOlszewski current blocker is to figure out what's wrong with those failing tests. Would you be interested in helping out with that?

PR #169, opened 2016-08-22
Update readme

GendelfLugansk:
Today I needed to remove 'Bearer ' from header, so I tried to change authorizationPrefix to empty string ('') and this didn't work. Looking into sources helped me, but I think this case should be described in readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,8 @@ ENV['ember-simple-auth-token'] = {
 };
 ```
 
+If you need to completeley remove authorization prefix ('Bearer: '), set it to null.
+
 For the JWT authenticator (in addition to the Token authenticator fields):
 
 ```


cloke:
Would be great to get this merged in. I had the same issue and fumbled around for an hour before discovering the same thing.

alechirsch (approved):

alechirsch:
@jpadilla Any idea why the tests are failing on this?

fenichelar:
@GendelfLugansk Great point, this was quite confusing. The latest change behaves as expected with a blank prefix. The `authorizationPrefix` should always be a string.

```js
this.authorizationPrefix = conf.authorizationPrefix === '' ? '' : conf.authorizationPrefix || 'Bearer ';
```

PR #182, opened 2016-12-21
Removed timeFactor config option and fixed refresh time calculation

Subtletree:
Expiration calculation is now only allowed in seconds, adhering to the
jwt spec.

### DIFF
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -125,7 +125,7 @@ export default TokenAuthenticator.extend({
       }
 
       if (expiresAt > now) {
-        const wait = expiresAt - now - (this.refreshLeeway * 1000);
+        const wait = (expiresAt - now - this.refreshLeeway) * 1000;
 
         if (wait > 0) {
           if (this.refreshAccessTokens) {
@@ -197,7 +197,7 @@ export default TokenAuthenticator.extend({
       expiresAt = this.resolveTime(expiresAt);
 
       const now = this.getCurrentTime();
-      const wait = expiresAt - now - (this.refreshLeeway * 1000);
+      const wait = (expiresAt - now - this.refreshLeeway) * 1000;
 
       if (!Ember.isEmpty(token) && !Ember.isEmpty(expiresAt) && wait > 0) {
         Ember.run.cancel(this._refreshTokenTimeout);
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -59,14 +59,6 @@ export default TokenAuthenticator.extend({
   tokenExpireName: 'exp',
 
   /**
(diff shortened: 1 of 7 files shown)


addon/authenticators/jwt.js
```diff
@@ -118,14 +109,14 @@ export default TokenAuthenticator.extend({
...
-        expiresAt = this.resolveTime(tokenData[this.tokenExpireName]);
+        expiresAt = tokenData[this.tokenExpireName];
         if (Ember.isEmpty(expiresAt)) {
           return resolve(data);
         }
       }
 
       if (expiresAt > now) {
-        const wait = expiresAt - now - (this.refreshLeeway * 1000);
+        const wait = (expiresAt - now - this.refreshLeeway) * 1000;
```
jpadilla, line 119: Why the change in parenthesis order?

iamsebastian:
Due to the fact, that the calculation of how-long-to-wait-before-token-gets-refreshed is imho broken, would be nice, if anybody could merge this.

As this does not makes any sense to me:
https://github.com/jpadilla/ember-simple-auth-token/blob/dd3cda1a6dd6d81a10ed919bf9b6a14fce72af32/addon/authenticators/jwt.js#L200

If I would `hack` the config of my ember-app, and set the `refreshLeeway` to `0.015` for testing out the refresh of the token, the calculated `wait` will force to call the refresh within some milliseconds.

screenshot from 2017-01-17 11-58-50

jpadilla:
@iamsebastian does this PR fix your issue?

iamsebastian:
@jpadilla It looks alike, that this could fix it. (The calculation) Needs to be tested tomorrow - by me. Maybe I install the patch and report back.

iamsebastian:
@jpadilla I've installed the patch via npm/GitHub dependency of @Subtletree 

Looks quite good. Would be great, if you could merge this:
screenshot from 2017-01-17 15-07-06


Thanks in advance

jpadilla (approved):

jpadilla:
@Subtletree @iamsebastian thanks! Just released this in v2.1.0

addon/authenticators/jwt.js
Subtletree: `wait` needs to be in milliseconds for `Ember.run.later`.

Previously `expiresAt` and `now` could be in either seconds or milliseconds which would make the `wait` calculation incorrect if either of them were in seconds.

Now `expiresAt`, `now` and `refreshLeeway` are all in seconds so get combined before being converted to milliseconds.

iamsebastian: @Subtletree is correct here.

iamsebastian:
Thanks a bunch, @Subtletree and @jpadilla

PR #183, opened 2016-12-22
Application adapter example updated

danwetherald:
This makes the documentation more clear to what you should be using as the authorizer in the application adapter once this addon is installed.


jpadilla:
@dan003400 thanks!

Issue #162, opened 2016-05-25
how does refreshing a token make my app secure?


jonnedeprez:
I may be missing something, but It's not clear to me how the current implementation of refreshing tokens improves the security of an application. I thought the reason to let a token expire was to prevent an intruder that somehow managed to intercept your access token from stealing your identity _indefinitely_. But if that intruder can use the same access token to simply refresh it on another endpoint before it expires, the protection can easily be bypassed. The only advantage here is that if a lazy intruder waits for a couple of weeks before he attempts to break in, he will fail! Not very useful, I'd say...
Wouldn't a better strategy be to use 2 different tokens? An access token that is used to access protected resources but that expires frequently, and a refresh token that never or almost never expires, but is only used to refresh the access token (as in [https://auth0.com/docs/refresh-token]).

eluciano11:
Hi, this library contains two implementations one of the if for simple token authentication (this one doesn't do any refreshes) and a JWT implementation. The blogpost you're referring to explains JWT and our current implementation in `jwt-authenticator` does the strategy you're suggesting (https://github.com/jpadilla/ember-simple-auth-token/blob/master/addon/authenticators/jwt.js). The support for simple token authentication will be dropped soon (https://github.com/jpadilla/ember-simple-auth-token/issues/151).

Hope this helps.
I'm closing this issue for now because it's not a library issue.

jonnedeprez:
Actually, my point was that you _claim_ that you implement that strategy in jwt.js, but that in reality you don't (no offense!). Your implementation of the JWT authenticator can indeed refresh the token, but it does so using the token as configured by tokenPropertyName (i.e. the access token). IMHO, this refreshing strategy doesn't improve security, because it also allows the token thief to call refresh.

evoactivity:
@eluciano11 I see this is an old issue, and was closed, but as @jonnedeprez restated, that is not what is currently implemented. My API has JWT implemented to give a short expiring JWT and a refresh token. Right now this library tries to use the JWT not the refresh token.

eluciano11:
Yeah, after looking more into it @jonnedeprez was right. The tokens are getting mixed up. I haven't had time to work on this but if you have time to fix this we will welcome you PR.

evoactivity:
@eluciano11 I've already forked ;) I should be able to have a look over the next few days.

eluciano11:
@evoactivity awesome, thank you!

jonnedeprez:
[...]

PR #190, opened 2017-02-13
Update refresh strategy to use a separate refresh token

evoactivity:
As discussed in this issue #162 I've made the library use a separate token to refresh the JWT.

I've updated the tests and they all pass but someone should really check what I've done, testing is not my strong point.

I've ran this on my app and confirmed it refreshes the token as expected.

### DIFF
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -68,6 +68,7 @@ export default TokenAuthenticator.extend({
     this.identificationField = Configuration.identificationField;
     this.passwordField = Configuration.passwordField;
     this.tokenPropertyName = Configuration.tokenPropertyName;
+    this.refreshTokenPropertyName = Configuration.refreshTokenPropertyName;
     this.refreshAccessTokens = Configuration.refreshAccessTokens;
     this.refreshLeeway = Configuration.refreshLeeway;
     this.tokenExpireName = Configuration.tokenExpireName;
@@ -98,6 +99,7 @@ export default TokenAuthenticator.extend({
     return new Ember.RSVP.Promise((resolve, reject) => {
       const now = this.getCurrentTime();
       const token = dataObject.get(this.tokenPropertyName);
+      const refreshToken = dataObject.get(this.refreshTokenPropertyName);
       let expiresAt = dataObject.get(this.tokenExpireName);
 
       if (Ember.isEmpty(token)) {
@@ -120,11 +122,11 @@ export default TokenAuthenticator.extend({
 
         if (wait > 0) {
           if (this.refreshAccessTokens) {
-            this.scheduleAccessTokenRefresh(dataObject.get(this.tokenExpireName), token);
+            this.scheduleAccessTokenRefresh(dataObject.get(this.tokenExpireName), refreshToken);
           }
           resolve(data);
         } else if (this.refreshAccessTokens) {
-          resolve(this.refreshAccessToken(token));
+          resolve(this.refreshAccessToken(refreshToken));
         } else {
           reject(new Error('unable to refresh token'));
         }
@@ -183,16 +185,16 @@ export default TokenAuthenticator.extend({
     @method scheduleAccessTokenRefresh
     @private
   */
-  scheduleAccessTokenRefresh(expiresAt, token) {
+  scheduleAccessTokenRefresh(expiresAt, refreshToken) {
     if (this.refreshAccessTokens) {
 
       const now = this.getCurrentTime();
       const wait = (expiresAt - now - this.refreshLeeway) * 1000;
 
-      if (!Ember.isEmpty(token) && !Ember.isEmpty(expiresAt) && wait > 0) {
+      if (!Ember.isEmpty(refreshToken) && !Ember.isEmpty(expiresAt) && wait > 0) {
         Ember.run.cancel(this._refreshTokenTimeout);
         delete this._refreshTokenTimeout;
-        this._refreshTokenTimeout = Ember.run.later(this, this.refreshAccessToken, token, wait);
+        this._refreshTokenTimeout = Ember.run.later(this, this.refreshAccessToken, refreshToken, wait);
       }
     }
   },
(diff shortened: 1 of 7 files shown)


eluciano11:
@evoactivity thank you for your work! 

I've looked at the test and the implementation and it looks good. I think that we should wait until @jpadilla can take a look at this before we merge the PR. This is a breaking change so we'll need to update the docs and demo too. We can do that after @jpadilla takes a look.

evoactivity:
@eluciano11 No problem :)

I was wondering about the docs/demos, but decided against updating them until this had been reviewed incase there are any changes needed.

addon/authenticators/jwt.js
```diff
@@ -120,11 +122,11 @@ export default TokenAuthenticator.extend({
 
         if (wait > 0) {
```
alechirsch: Don't we want to check if the refresh token has expired rather than if the authentication token has expired? That is my understanding of refresh tokens, they out last the authentication tokens.

addon/authenticators/jwt.js
```diff
@@ -120,11 +122,11 @@ export default TokenAuthenticator.extend({
```
alechirsch, line 122:
```javascript
if (Ember.isEmpty(expiresAt)) {
         // Fetch the expire time from the token data since `expiresAt`
         // wasn't included in the data object that was passed in.
         const tokenData = this.getTokenData(token);
 
         expiresAt = tokenData[this.tokenExpireName];
         if (Ember.isEmpty(expiresAt)) {
           return resolve(data);
         }
       }
 
       if (expiresAt > now) {
```

Shouldn't we be checking if the refresh token `expiresAt > now` rather than the auth token `expiresAt > now`?

With the way it currently is, if the auth token expired but the refresh token has not expired, it would not attempt to refresh.
I think we would want to attempt to refresh if the refresh token is still valid.

evoactivity: My API's refresh token library creates the token like so `$this->refreshToken = bin2hex(openssl_random_pseudo_bytes(64));`, so the refresh token is not just another JWT with an expires property to be decoded. Looking at the auth0 page on refresh tokens it looks like they're implementing them in a similar way https://auth0.com/learn/refresh-tokens/

If the auth token is expired that's exactly when this tries to refresh using the refresh token, so I'm not entirely sure what you mean by that?

alechirsch: From what I am seeing, if the auth token is expired when it enters the restore function then it will reject on line 132
```
 reject(new Error('token is expired'));
```

evoactivity: Ahh, of course I see what you mean now. Yeah, you're right, if your token is already expired it won't be refreshed.

I think the change here would be if it's expired to attempt a refresh, if the server rejects the refresh token then unauth the session so they are sent to the login form.

evoactivity: Oh yes, I see what you mean now, my mistake.

I guess if the token is expired and we have `refreshAccessTokens` set to true it should also attempt to refresh the token there as well, and if the server rejects the refresh token then the user is unauthorised sending them back to the login form.

alechirsch: Yes, sorry for the confusion.

jpadilla (approved):

jpadilla:
@evoactivity At a first pass, this lgtm. Thanks for putting this together.
@eluciano11 is there anything we'd need after this, docs and dummy app?

I'm fine introducing this as a new major release, but would also love to sneak in updating everything else including to the latest ESA.

addon/authenticators/jwt.js
alechirsch: @evoactivity Do you have time to implement the fix for this? I am considering forking off, I will be trying to use this feature in an app for this week.

evoactivity: @alechirsch I will be looking at this again maybe later today or tomorrow.

evoactivity: @alechirsch I've updated this to attempt a refresh if the JWT is expired. If the server rejects the request the session is invalidated.

alechirsch: @evoactivity Awesome, thanks for your work on this, it is perfect timing.

evoactivity:
@jpadilla no worries :) I've just pushed a small update to address the expiry issue pointed out by @alechirsch 

The docs will need updating to explain the new `refreshTokenPropertyName` option, if people want to use the current strategy they should be able to simply give that the same name as `tokenPropertyName`.

The demo app should probably be updated to show this being used, the demo server will need to return two tokens now, the JWT and a standard refresh token.

alechirsch (approved):

eluciano11:
@jpadilla @evoactivity yup, updated the docs and the demo app but it's not using the two tokens it's falling back to the old strategy.

alechirsch:
I just did some testing in my own application, there seems to be an issue with old timeouts not being removed. I will do more testing to find out what is really going on.

alechirsch:
@evoactivity From what I am finding if you log in then refresh the page to make it call restore, it will refresh the token, and then call restore again exponentially. Meaning it calls restore once, then twice, then 4 times, then 8. Something about `resolve(this.refreshAccessToken(refreshToken));` is making it call restore twice for every time it is called.

Restore should be called only once, when the page is refresh or opened in a new tab. I am unsure why this is happening at the moment.

evoactivity:
@alechirsch I've not seen that behaviour at all in my own app.

What sort of expiry time do you have on your JWT? Also, what leeway do you have set?

alechirsch:
For testing purposes I set it to 60 seconds, with 55 second leeway. Maybe that is the problem?

evoactivity:
I've been testing with my JWT set to 1 minute and my refresh token set to 2 minutes and my leeway set to 0.

I'm not sure why a 5 second expiry would call exponential calls to refreshAccessToken though.

alechirsch:
What's odd is that the restore function is being called more than once.

alechirsch:
I blew away my node_modules and reinstalled, set my expire to 10 minutes and leeway to 60 seconds. I am running into the same problem, the restore function is being called in an infinite loop somehow. It works great if the restore function rejects the first time it is hit, however if it resolves, it calls itself again. It hogs up the CPU at 100% as well.

alechirsch:
I have traced it back to this file `ember-simple-auth/instance-initializers/setup-session-restoration`.

The beforeModel hook is being called on an infinite loop, which is in turn calling restore. This might be a problem with my app, looking into it more.

alechirsch:
@evoactivity It was definitely a problem with my application, the feature works as expected now.

jpadilla:
This is the current server implementation that the dummy app uses https://github.com/jpadilla/ember-simple-auth-token-server

alechirsch:
What is currently holding up this PR? I have been using this for the past few weeks in my apps and it works perfectly by the way.

jpadilla:
@alechirsch thanks for bumping this again, great to know you've been using it for a while now. Since this is a breaking change, it'll be released on v3, but before doing so I would love to make sure we update ember-cli, ESA, etc. Thanks to all!

PS If anyone else wants to help with that before I do, please feel free to do so. Thanks!

jpadilla:
@homu r+

jpadilla:
@alechirsch @evoactivity I'm looking for help with maintenance, would any of you be up to it?

alechirsch:
@jpadilla I can help out with pull requests and issues

jpadilla:
@alechirsch that's so great, just sent you an invite, thank you so much!

evoactivity:
@jpadilla I'm too busy to offer much help at the moment but I'd be interested in helping when things settle down a bit for me. That's probably a month or two away though.

jpadilla:
@evoactivity understandable! Feel free to ping back whenever you have a few cycles free, thanks!

knownasilya:
This is a breaking change, because if the refreshToken name isn't set it defaults to `refresh_token`, which used to be the token name. If the property isn't set, it should default to the token name.

jpadilla:
Thank you all again! Just released [v3.0.0](https://github.com/jpadilla/ember-simple-auth-token/releases/tag/v3.0.0) 🎉

PR #196, opened 2017-04-04
Add optional one time password support

fenichelar:


fenichelar:
@jpadilla Thoughts?

alechirsch:
@fenichelar Can you explain the use case for this? I do not see why a user cannot use the password field for a one time password.

fenichelar:
This is what my login screen looks like: 
image

Generally you use both a password and a one time password. GitHub, Facebook, Amazon, etc. all do it this way.

alechirsch:
I think the documentation could use some more detailed explanation of what "otp" is for, It can be vague and confusing for some users

jpadilla:
@fenichelar thanks for putting this together. I'm -1 for adding this for now, because otp can vary a lot between implementations(ex: ask for otp after login).

If someone needs to do something like this, they'd only need to extend our authenticator and add their field in `getAuthenticateData()`.

@alechirsch thoughts?

alechirsch:
@jpadilla I definitely agree that otp can vary, and it might not be sent along in the same request as a username and password.

I have actually needed to add my own field to the authentication request and found that extending the authenticator was the way to go.

```javascript
import JWT from 'ember-simple-auth-token/authenticators/jwt';

export default JWT.extend({
	getAuthenticateData(credentials){
		return credentials;
	},
});
```
You can also call this._super and add 'otp' to the credentials, but I found this to be sufficient.

I do not think that 'otp' necessarily belongs in the plugin if you can easily add your own customized fields.

fenichelar:
This is definitely something that should be an extension. Some documentation in the readme about extending the authenticator would be helpful.

Issue #50, opened 2015-04-28
Return data record from authentication


Baukereg:
I would like to return a data record from the authenticate request. Something like this:

``` javascript
{
   "token":"eyJhbGciOiJIUzI1NiIsIskdj........",
   "account":{
      "id":"account_1",
      "username":"Demo",
      "role":"CLIENT",
      "personKey":"client_1"
   }
}
```

Account should be deserialized to an account model record.

Is there any way to accomplish this in the current version of simple auth token?

injaon:
I have to deal with the same issue today. This is how I resolved it.

First, I modify my backend server (django) to return the user id with the token. So, with a succesfull authentication it returns something like this:

``` javascript
{
   "token": "eyJhbGciOiJIUzI1NiIsIskdj........",
   "empleado_id": 1234
}
```

Then, I created a custom session. When it receives the token it makes a secord request for the user data. This is the code:

``` javascript
/* app/initializers/crystal-session.js */

import Em from "ember";
import DS from "ember-data";
import Session from "simple-auth/session";

var CrystalSession = Session.extend({
    user: function() {
        var accountId = this.get('empleado_id');
        if (!Em.isEmpty(accountId)) {
            return DS.PromiseObject.create({
                promise: this.container.lookup('store:main').find('empleado', accountId)
            });
        }
        return null;
    }.property('empleado_id')
});

export default {
    name: 'crystal-session',
    before: "simple-auth",
    initialize: function (container, application) {
        application.register('session:crystal', CrystalSession);
    }
};
```

**Please notice** that my user is called _empleado_

Finally, let simple-auth know about this session. Edit your `config/environment.js` file with:

``` javascript
ENV['simple-auth'] = {
    session: 'session:crystal',
    /* other stuff here */   
}
```

aibarra11:
Hello Injaon, Can you please show your controller code, particular the authenticate action that's triggered when submitting the username/password.  I followed your example, but the user function isn't being triggered, so I'm wondering if you're using a regular ajax call or this.get('session').authenticate(authenticator, credentials) method in this package.

thermokarst:
@aibarra11: I had implemented something based on @injaon's response, however I just realized that for what I am trying to do, this was probably overkill. I just reimplemented to something along these lines (see below). I completely removed the custom session initializer, and did the following in my login controller:

``` javascript
// app/pods/login/controller.js
import Ember from 'ember';
import parseBase64 from '../../utils/parse-base64';

[...]

Issue #120, opened 2016-01-13
Cannot get custom data from session.data attribute


danr1979:
Hi.

I'm running ember-simple-auth 1.0.1, ember-simple-auth-token 1.0.0 and Ember 2.2.0 with Ember Data 2.2.0.

I've got things working and I'm able to authenticate my Ember app with the server-side PHP API backend running Firebase PHP-JWT. When generating my token on the server side to send back to ember-simple-auth I pass in a "data" attribute containing a "userId" and a few other custom values as follows:

```
$token = array(
    'iss' => env('HOST_NAME'),
    'aud' => env('HOST_NAME'),
    'exp' => $expires,
    'data' => array(
        'userId' => $user->id,
        'roles' => $roles,
        'permissions' => $permissions
    )
);

$jwt = JWT::encode($token, $key, 'HS512');
```

However, in my Ember app if I try to access this data after authenticating like so:

```
this.get('session.data')
```

all I see is the default "authenticated" key with the "exp" and "token" etc inside. How do I go about accessing my custom data values ("userId", "roles" etc)? I've read the documentation but I can't find an example of the data being access, only set.

The response the server is returning when authentication succeeds is as follows:

{"token":"eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzUxMiJ9.eyJpc3MiOiJodHRwczpcL1wvZGV2LnNlY3VyZS1vcGVyYXRpb25zLmNvbSIsImF1ZCI6Imh0dHBzOlwvXC9kZXYuc2VjdXJlLW9wZXJhdGlvbnMuY29tIiwiZXhwIjoxNDUyNjcyODY2LCJkYXRhIjp7InVzZXJJZCI6IjUiLCJyb2xlcyI6W10sInBlcm1pc3Npb25zIjpbXX19.ivO289WrKI2q-HBLhiHm5tiYZSeqKr1yPueddjW5gZD2s986OaR62sr7fA-Y7tonzj4hmk9H5yCjY1f_hjK2MQ"}

This appears to be a valid token and, interestingly, when the token is passed back to the server in subsequent requests to other API resources I can pick up the "data" fields from it in PHP. This includes my custom fields so they are present in the token but appear to not be being recognized on the client side.

Hopefully this helps to understand what is amiss?

I'd appreciate any help anyone can offer.

Thanks.

danr1979:
[...]

PR #206, opened 2017-08-31
Make token data available to session

tomclose:
This PR changes the jwt authenticator to include all the fields in the `tokenData` in the `sessionData` it provides, so that the contents of the token will be available on the ember session.

For example, if your token payload is the following:
```
{
  "sub": "1234567890",
  "name": "John Doe",
  "admin": true
}
```
then you'll be able to do the following:
```
session.get('data.authenticated.name') //=> 'John Doe'
```

This fixes issues https://github.com/jpadilla/ember-simple-auth-token/issues/50 and https://github.com/jpadilla/ember-simple-auth-token/issues/120.

This PR includes https://github.com/jpadilla/ember-simple-auth-token/pull/207.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
-  - "0.12"
+  - "8.4.0"
 
 sudo: false
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,4 +31,4 @@ install:
   - bower install
 
 script:
-  - ember try:each
+  - ember try:one $EMBER_TRY_SCENARIO
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -369,7 +369,7 @@ export default TokenAuthenticator.extend({
       this.scheduleAccessTokenRefresh(expiresAt, refreshToken);
     }
 
-    return assign(this.getResponseData(response), tokenExpireData);
+    return assign(this.getResponseData(response), tokenData, tokenExpireData);
   },
 
   /**
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,9739 @@
+{
+  "name": "ember-simple-auth-token",
+  "version": "2.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "abbrev": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.0.tgz",
+      "integrity": "sha1-0FVMIlZjbi9W58LlrRg/hZQo2B8=",
+      "dev": true
+    },
+    "accepts": {
+      "version": "1.3.4",
+      "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.4.tgz",
+      "integrity": "sha1-hiRnWMfdbSGmR0/whKR0DsBesh8=",
+      "dev": true,
+      "requires": {
+        "mime-types": "2.1.16",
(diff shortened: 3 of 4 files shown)


tomclose:
@jpadilla does this seem like a reasonable thing to do? Alternatively, is there another way you'd recommend to get at the contents of the token?

ghost:
@tomclose, Why you need to overload your jwt payload with these information? You can't use another API for "current_user"? I think the payload should be as light as possible because it is sent back and forth every time. Or am I wrong, @jpadilla?

tomclose:
@johnunclesam as far as I know, this is one of the advantages to using a jwt. From [jwt.io](https://jwt.io/introduction/):

> JSON Web Token (JWT) is an open standard (RFC 7519) that defines a compact and self-contained way for securely transmitting information between parties as a JSON object. 

and

> Self-contained: The payload contains all the required information about the user, avoiding the need to query the database more than once.

They then give the following example payload:
```
{
  "sub": "1234567890",
  "name": "John Doe",
  "admin": true
}
```
You definitely could use another api call to fetch the current user, but it seems like pulling it out of the jwt payload is also a valid thing to do.

tomclose:
Also worth saying: in my actual use-case, the piece of information I want to extract from the token is just the `user_id`, not the entire user.

kcaulfield94:
+1

jherdman:
This would be really great. I'd love to use it. Is there anything I can do to help to get this moving?

jherdman:
Actually, I do have a question for you. Is this resilient to page refreshes? To me it looks like we're setting this data once we do an authentication request, but the data would be missing upon a page refresh. Perhaps we need to read the data out of the session storage service instead.

gzurbach:
Having my user ID in the JWT payload I'd love to be able to access token data directly from the session service. Hope this change moves forward in future versions.

donaldwasserman:
I'd actually like to chime in with a bit of a counter point on this topic. I don't think this addon should be adding data into the session without some sort of extra work by application. 

I totally see the use case, and I actually do quite a bit of it through a custom authenticator. There's some data we send in our JWT's that we don't want persisted in the session (ie. last login time, total site visits). 

So a better approach to this PR, I think would be having at least some sort of configuration flag that either include/exclude certain keys.

fenichelar:
@donaldwasserman "There's some data we send in our JWT's that we don't want persisted in the session" - is there a reason it shouldn't be persisted? If the token contains the data, the data will be persisted in the token. It is just a matter of persisting it decoded in addition to persisting it encoded.

donaldwasserman:
@fenichelar - I just save some random properties I don't necessarily want automatically overridden by refresh tokens and whatnot. If I'm the only person who does this, I'm not going to pitch a fit about it and will update my app.

fenichelar:
@donaldwasserman Does it make sense to put the decoded properties in a object like `tokenData` so it won't override any other things set by the application?

donaldwasserman:
@fenichelar I think that's a good idea. 

The Ember Auth0 addon uses a similar strategy: https://github.com/auth0-community/ember-simple-auth-auth0#data-object-on-the-session

Doing that would give some people at least a similar point of reference between the two JWT authentication.

fenichelar:
@tomclose Great idea! Based on your idea, I made the token data available in the session but I made it available through `tokenData` instead of being available directly in the session:

```js
session.get('data.authenticated.tokenData.name') //=> 'John Doe'
```

I think this is a good approach based on @donaldwasserman's concerns.

fenichelar:
I think the new solution solves this so unless there are any other concerns, I'll close :)

PR #207, opened 2017-09-03
Fix tests

tomclose:
This PR gets the tests passing on master again. To do this it updates node and adds a lockfile, which avoids the issue with incompatible libraries [here](https://travis-ci.org/jpadilla/ember-simple-auth-token/jobs/233670467).

It also changes the call to `ember try`, as currently every scenario was being run on every travis job, instead of just the scenario for that job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 ---
 language: node_js
 node_js:
-  - "0.12"
+  - "8.4.0"
 
 sudo: false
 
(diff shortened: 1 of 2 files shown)


tomclose:
@jpadilla is this an approach that you're happy with?

jpadilla:
Sorry for the delay here, this should be resolved now with the latest release. Thanks again!

Issue #216, opened 2018-02-02
JWT payload is not decoded properly


abhishek97:
I was just trying to implement a jwt token auth with refresh token for my app, however, i was getting this error when ember-simple-auth-token tries to decode the jwt received from the server

`Failed to execute 'atob' on 'Window': The string to be decoded is not correctly encoded.`

I looked into the issue for a bit, tracked it down to this line: https://github.com/jpadilla/ember-simple-auth-token/blob/master/addon/authenticators/jwt.js#L284

The thing is we try to do a base64 decode, however as specified in the JWT rfc7519(https://tools.ietf.org/html/rfc7519), the payload is base64url encoded, so `window.atob` will fail to decode it.

abhishek97:
I seemed to have fixed this, I have also submitted the PR for the same. If someone can look into it please

PR #217, opened 2018-02-02
base64url decode jwt

abhishek97:
https://github.com/jpadilla/ember-simple-auth-token/issues/216


maxwondercorn:
You need to modify .travis.yml with the following to eliminate the sandbox errors and pass the tests:

sudo: required
dist: trusty

addons:
  chrome: stable
before_script:
    - "sudo chown root /opt/google/chrome/chrome-sandbox"
    - "sudo chmod 4755 /opt/google/chrome/chrome-sandbox"

Issue #218, opened 2018-04-03
ESA Authorizer deprecation


scottkidder:
Authorizers are deprecated and will be removed in ESA 2.0
https://github.com/simplabs/ember-simple-auth#deprecation-of-authorizers

jpadilla:
Closed by #226

PR #219, opened 2018-04-14
Add TokenAuthorizerMixin

sukima:
This fixes issue #218

All options are the same as the old authorizer but uses the new syntax. The Mixin will import DataAdapterMixin and session for you.

I was unsure on tests so would appreciate anyone pushing a commit or two to help with that.

Added a blurb to the README but it could use some help from someone more articulate then I.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,18 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
 ```
 *DataAdapterMixin is supported for Ember 1.13 and above*
 
+### Deprecation in ember-simple-auth 2.0
+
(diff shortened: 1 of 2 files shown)


sukima:
I did not touch `yarn` which is the command failing on CI. Therefor **CI is lying!**

jpadilla:
Started working on updating to latest ember-cli and hopefully making sure all tests are passing again. Any help there is more than welcome. https://github.com/jpadilla/ember-simple-auth-token/pull/222

jpadilla:
Closed via #226. Your original commit was preserved. Thanks again!

PR #224, opened 2018-05-12
deprecate `getAuthenticateData`

### LINKED ISSUE #223: Proposal: Remove Use of `getAuthenticateData`
While I understand the value of `getAuthenticateData` as a way of sanitizing data, it does cause a lot of confusion and misdirection for users while providing little in terms of extensibility.

https://github.com/jpadilla/ember-simple-auth-token/blob/master/addon/authenticators/token.js#L146-L153

Also it does not use `Ember.get` so you could lose the potential value of unwrapping objects, proxies or computed properties.
This is something that could have been useful in a sanitization/unwrapping function.

Passing `credentials` directly as an object would be clearer from a use perspective and would really slim down confusion around `identificationField` and `passwordField` config since they wouldn't be needed anymore.
This also makes the library more flexible for use with other systems which may require more than two fields.
I have worked in projects where other session IDs are passed in along with email/password to the token endpoint and this would require extending and rewriting the `getAuthenticateData` or even the entire `makeRequest` function.

---

Example in an app may have a controller with the following code:

```
this.session.authenticate('authenticator:token', { identification: this.email, password: this.password })
```

If after API development the JWT endpoint requires `email` and `password` instead of the configured `identification` and `password` then the thought would be to change the code to:

```
this.session.authenticate('authenticator:token', { email: this.email, password: this.password })
```

But instead the configuration for `identificationField` would need to be changed to `email`.

---

In comparison if there was no `getAuthenticateData` function or call, then it becomes much more clear and simple to pass in the fields required by the API being used.
Th code change listed be required and users wouldn't have to know about configuration for `identificationField` or `passwordField` at all, and they would have more flexibility.

---

Downside:

Removing `getAuthenticateData` would potentially break applications where credentials aren't explicitly passed in (for instance using `this` as the credentials in a controller/component and letting `getAuthenticateData` remove properties other than the required fields).
This would be hard to make a warning for.

The change required to support this in an app would be to either explicitly create a new credentials object (I think this is best to recommend) or using something like `_.pick` or `Ember.Object.getProperties`.

rtablada:
This closes #223 as a way to deprecate `getAuthenticateData` in v3.

It does this by checking if the config is set to something that is not default or if the `credential` object has more than 2 keys (which was historically stripped down to only `identification` and `password` by `getAuthenticateData`.

In the next version, the `if` check and `getAuthenticateData` would be removed entirely.

### DIFF
--- a/addon/authenticators/token.js
+++ b/addon/authenticators/token.js
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import Base from 'ember-simple-auth/authenticators/base';
 import Configuration from '../configuration';
+import { deprecate } from '@ember/application/deprecations';
 
 /**
   Authenticator that works with token-based authentication like JWT.
@@ -125,7 +126,11 @@ export default Base.extend({
   */
   authenticate(credentials, headers) {
     return new Ember.RSVP.Promise((resolve, reject) => {
-      const data = this.getAuthenticateData(credentials);
+      let data = credentials;
+
+      if (this.checkCredentialGetAuthData(credentials)) {
+        data = this.getAuthenticateData(credentials)
+      }
 
       this.makeRequest(data, headers).then(response => {
         Ember.run(() => {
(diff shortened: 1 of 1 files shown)


jpadilla:
Closed via #226. We decided to remove `getAuthenticateData` now as part of the next major version. Thanks again!

PR #195, opened 2017-03-31
Invalidate session after token expiration

jembezmamy:
Here's mine attempt to resolve #174. I've added `handleAccessTokenExpiration` method that invalidates session as soon as token expires. I had to change some existing tests since they were just checking if `Ember.run.later` was called and now I use it also for scheduling expiration.

PR #225, opened 2018-05-14
Update Ember, redesign configuration, add features

fenichelar:
This includes changes from the following PRs:

- https://github.com/jpadilla/ember-simple-auth-token/pull/224
- https://github.com/jpadilla/ember-simple-auth-token/pull/222
- https://github.com/jpadilla/ember-simple-auth-token/pull/219
- https://github.com/jpadilla/ember-simple-auth-token/pull/195

Simplified environment configuration loading.

Also, switched from yarn to npm to fix Travis issues. Can be switched back.

Version 4.0.0 because it includes breaking changes - remove `getAuthenticateData ` (https://github.com/jpadilla/ember-simple-auth-token/pull/224).

### DIFF
--- a/addon/authenticators/token.js
+++ b/addon/authenticators/token.js
@@ -145,8 +145,8 @@ export default Base.extend({
   */
   getAuthenticateData(credentials) {
     const authentication = {
-      [this.passwordField]: credentials.password,
-      [this.identificationField]: credentials.identification
+      [this.passwordField]: credentials[this.passwordField],
+      [this.identificationField]: credentials[this.identificationField]
     };
 
     return authentication;


fenichelar:
Well it took a few tries but I got it working! :)

@jpadilla Awaiting your feedback.

jpadilla:
@fenichelar wow, thanks for this. I still haven't finished fully reviewing this, and might not have the time to do so completely, but I won't let that stop this from moving forward. If anyone else wants to checkout this pull request and try it on their app and report back that'd be awesome.

On another note, how crazy would it be to actually rebase changes from other pull requests into this one? That way we're able to give credit where it's due. After rebasing you'd have to fix some things around, but their commits will be kept.

I'd also prefer to keep `getAuthenticateData` around and later merge #224 so we can cut out a release before jumping to v4.

Thanks again for all your hard work! 🎉 🎆

fenichelar:
> On another note, how crazy would it be to actually rebase changes from other pull requests into this one? That way we're able to give credit where it's due. After rebasing you'd have to fix some things around, but their commits will be kept.

Done. New PR inbound.

> I'd also prefer to keep getAuthenticateData around and later merge #224 so we can cut out a release before jumping to v4.

A major part of the broken tests and demo were caused by using an incorrect `identificationField`. The removal of `identificationField` and `passwordField` and in turn `getAuthenticateData` are a major simplification. Frankly, I'm not sure I want to spend the effort to get it working with these features if the plan is to remove it later. Also, upgrading from Ember 2.x to 3.x seems like the perfect time to upgrade from 3.x to 4.x.

fenichelar:
https://github.com/jpadilla/ember-simple-auth-token/pull/226

PR #200, opened 2017-05-18
fix: Use ember-fetch to avoid test failures

knownasilya:
closes #187

I'm not sure how to stub fetch, and fix the tests. Any guidance is appreciated.

Maybe something like http://www.wheresrhys.co.uk/fetch-mock/ ?

PR #188, opened 2017-01-25
Fixes https://github.com/jpadilla/ember-simple-auth-token/issues/170

poohitan:
This is duplicate of #173 with fix of #170 by @Eusse, I've just fixed merge conflicts. Hope it will help this fix to be merged faster.

PR #187, opened 2017-01-24
Fixed jquery 3 promise handling to be compatible with ember promises

pfmmfp:
More info explained [here](http://jquery.com/upgrade-guide/3.0/\#callback-invocation), which has been found based on [this ember issue](https://github.com/emberjs/ember.js/issues/14566)

PR #167, opened 2016-07-14
Change ember-simple-auth 1.0.1 to ^1.1.0

brendan-rius:
Should fix #166

PR #226, opened 2018-05-17
Update Ember, redesign configuration, add features

### LINKED ISSUE #170: identificationField can't overwrite
Trying to use `email` instead of the default `identification` with `identificationField: 'email'` but it's not working. It only works if I use `identification` in my login form. Any ideas?

fenichelar:
This includes changes from the following PRs:

- https://github.com/jpadilla/ember-simple-auth-token/pull/219 Add TokenAuthorizerMixin
- https://github.com/jpadilla/ember-simple-auth-token/pull/217 base64url decode jwt
- https://github.com/jpadilla/ember-simple-auth-token/pull/195 Invalidate session after token expiration

It also:

- Simplifies environment configuration loading
- Organizes testing
- Stores token data in the session
- Switches from yarn to npm
- Switches from `jquery` to `ember-fetch`
- Removes `getAuthenticateData`
- Removes `getResponseData`
- Uses `eslint-plugin-ember`
- Merges ajax request function in `JWT` and `token`
- Throws error if `refreshLeeway` is too large

In addition to closing the above PRs, the following PRs are no longer needed if accepted:

- https://github.com/jpadilla/ember-simple-auth-token/pull/224 deprecate `getAuthenticateData`
  - `getAuthenticateData ` removed 
- https://github.com/jpadilla/ember-simple-auth-token/pull/222 Update ember-cli to 3.1.2
  - `ember-cli` updated
- https://github.com/jpadilla/ember-simple-auth-token/pull/206 Make token data available to session
  - Token data available to the session
- https://github.com/jpadilla/ember-simple-auth-token/pull/200 fix: Use ember-fetch to avoid test failures
  - Tests fixed
- https://github.com/jpadilla/ember-simple-auth-token/pull/188 Fixes https://github.com/jpadilla/ember-simple-auth-token/issues/170
  - `getAuthenticateData ` removed
- https://github.com/jpadilla/ember-simple-auth-token/pull/187 Fixed jquery 3 promise handling to be compatible with ember promises
  - Switched to `ember-fetch`
- https://github.com/jpadilla/ember-simple-auth-token/pull/169 Update readme
  - Setting `authorizationPrefix` blank now works as expected
- https://github.com/jpadilla/ember-simple-auth-token/pull/167 Change ember-simple-auth 1.0.1 to ^1.1.0
  - `ember-simple-auth` updated
- https://github.com/jpadilla/ember-simple-auth-token/pull/122 Add warning when refreshLeeway is set to high
  - An error is thrown if `refreshLeeway` is too large

This would leave only the following PR left and I think the feature need discussion before implementing:

- https://github.com/jpadilla/ember-simple-auth-token/pull/152 Invalidate on idle functionality


fenichelar:
If accepted, I think it makes sense to address the following before releasing version 4.0.0:

- https://github.com/jpadilla/ember-simple-auth-token/pull/187
- https://github.com/jpadilla/ember-simple-auth-token/pull/122

This would leave only the following PRs left and I think both features need discussion before implementing:

- https://github.com/jpadilla/ember-simple-auth-token/pull/206
- https://github.com/jpadilla/ember-simple-auth-token/pull/152

jpadilla:
@fenichelar thanks! Fair enough, I don't mind jumping directly to v4 with some docs in place.

I'm fine with leaving #206 and #152 out of this. Regarding #187, what did you try? I suggested using ember-fetch https://github.com/jpadilla/ember-simple-auth-token/pull/187#issuecomment-296715795 although we weren't sure how testing would look like. The [DeviseAuthenticator unit tests](https://github.com/simplabs/ember-simple-auth/blob/master/tests/unit/authenticators/devise-test.js) use pretender.

fenichelar:
Based on your comment, I used ember-fetch. It worked including the demo. But I was having issues spying on the fetch function. I’ll post what I have in the comments when I get to my computer.

fenichelar:
It's a little too much too put in the comments. Maybe it makes more sense to focus on this PR and then focus on switching to `ember-fetch`. `ember-fetch` definitely works, the problems is the tests.

jpadilla (approved):
@fenichelar this is so good, thanks again for taking the time to work on this. I currently don't have an actual CHANGELOG for this project, usually just do notes on the actual Release. Any chance you could add a section to the README with some upgrade notes?

fenichelar:
@jpadilla Yes. I'll also try and improve the documentation about the authorizer.

fenichelar:
@jpadilla The readme has been updated. I also made some small code changes. Thoughts?

https://github.com/fenichelar/ember-simple-auth-token/blob/master/README.md

fenichelar:
I figured out how to get the `ember-fetch` tests working. I'm rewriting all the existing tests to get everything working, organized, and consistent. It might take a few days to get it finished.

fenichelar:
I made the changes as discussed, updated original post.

Summary:

- Organizes testing
- Stores token data in the session
- Switches from `jquery` to `ember-fetch`
- Removes `getResponseData`

I'm looking into the Travis build error. See https://github.com/ember-cli/broccoli-lint-eslint/issues/128

fenichelar:
So it looks like the Travis build issue were unrelated. Some dependency issue. Fixed now.

package.json
```diff
@@ -55,7 +55,7 @@
     "ember-resolver": "^4.0.0",
     "ember-simple-auth": "^1.6.0",
     "ember-sinon": "^2.1.0",
-    "ember-source": "~3.1.0",
+    "ember-source": "~2.16.0",
```
jpadilla: @fenichelar why the downgrade here?

fenichelar: I didn't mean to do it, I removed `node_modules` and ran `npm install` and it was done automatically. I'll fix.

fenichelar: I should have read the version number more carefully... lol

fenichelar: Fixed and Travis approves!

jpadilla (approved):
@fenichelar You've done such an amazing work here, thank you so much. There's maintainers spot with your name on it, if you're up for it!

fenichelar:
One more thing I am thinking about, I personally don't use separate refresh tokens. To refresh your token, you just use your access token and make a regular API request. Unfortunately, my backend has to support reading the token from the Authorization header and reading the token from the body. It would be great if refresh requests could be made with the refresh token in the Authorization header.

I was thinking of adding a config setting `refreshAccessTokenWithHeader` which defaults to false and the current behavior. If set to true, the refresh request would be made with the refresh token (which in my case is the same as the access token) in the `authorizationHeaderName` header instead of the body. The body would be an empty JSON object. The reason I think it should be empty is because I have to strip tokens from both places in my logging system. It would be great to just have one.

Thoughts? I could easily override the `refreshAccessToken` method for my projects if no one else would benefit.

fenichelar:
We use JWT with our Sails.js backend for most of our products. This library has been great! I'm happy to help maintain.

pusle:
I am very grateful for this addon, and for the work done on this deprecation fix. I have just upgraded from ember 2.x to ember 3.x, and would be very happy if I could test this branch. Alas, I am not very traversed in neither ember nor git, so if it is not to much work I would love to get the knowledge on how I can switch to this latest implementation and test it. Thanks.

fenichelar:
@pusle Great! Take a look at the readme [here](https://github.com/fenichelar/ember-simple-auth-token)

Instead of running:

```
ember install ember-simple-auth
ember install ember-simple-auth-token
```

Run:

```
ember install ember-simple-auth
npm install git://github.com/fenichelar/ember-simple-auth-token.git#master --save-dev
```

abhishek97:
@fenichelar Thanks so much for this. We use this across all of our products, but due to deadlines used our own fork. Would be great to join again!

fenichelar:
@abhishek97 Be aware there are a few breaking changes, this will be released as 4.0.0. Specifically, `getResponseData`, `getAuthenticateData`, `config.identificationField`, and `config.passwordField` have been removed.

fenichelar:
@jpadilla It may make sense to wait a bit before doing the npm release to allow more testing. I will deploy to production today and confirm we aren't having issues.

jpadilla:
> I was thinking of adding a config setting refreshAccessTokenWithHeader which defaults to false and the current behavior. If set to true, the refresh request would be made with the refresh token (which in my case is the same as the access token) in the authorizationHeaderName header instead of the body. The body would be an empty JSON object. The reason I think it should be empty is because I have to strip tokens from both places in my logging system. It would be great to just have one.

@fenichelar At this point I'm inclined to think we should just have an API that allows things like what you need to do without having to overwrite the whole authenticator/authorizer. Seems like a pretty specific use case.

pusle:
@fenichelar Thank you. I have tested on my local dev environment, and it went as smoothly as one could only dream of. I will push it to the dev server today, and to prod tomorrow evening.

Thank you again for your work on this fix!

Edit: I have pushed to the dev server, and no issues found by me or any one else.

fenichelar:
> At this point I'm inclined to think we should just have an API that allows things like what you need to do without having to overwrite the whole authenticator/authorizer. Seems like a pretty specific use case.

@jpadilla What do you mean? What I was going to do was just create a jwt authenticator that extends the current one and overwrite the `refreshAccessToken` function. It should be pretty simple:

```js
import { merge } from '@ember/polyfills';
import JWT from 'ember-simple-auth-token/authenticators/jwt';

export default JWT.extend({
  refreshAccessToken(token) {
    return new Promise((resolve, reject) => {
      const headers = merge(this.headers, {
        'Authorization': `Bearer ${token}`
      });
      this.makeRequest(this.serverTokenRefreshEndpoint, {}, headers).then(response => {
        try {
          const sessionData = this.handleAuthResponse(response);
          this.trigger('sessionDataUpdated', sessionData);
          return resolve(sessionData);
        } catch (error) {
          return reject(error);
        }
      }).catch(error => {
        this.handleTokenRefreshFail(error.status);
        return reject(error);
      });
    });
  }
});
```

fenichelar:
@pusle Great. I have not found any issues yet either. Do you use JWT and refresh tokens? Are the refresh tokens and access tokens separate?

fenichelar:
@jpadilla I'm thinking it might be time to publish version 4.0.0. I've had this in production for over a week now without issues (JWT). Thoughts?

pusle:
@fenichelar Sorry for the late reply. I use JWT through the Lexik JWT Authentication Bundle for Symfony. Refresh tokens, I believe, are separate. If I remember correctly.

fenichelar:
@jpadilla I am using the same token for both refresh and access. @pusle is using different tokens. I think that means most of the JWT class has been tested in real life. This combined with the unit tests means I think we are ready to release!

jpadilla:
@fenichelar just released v4.0.0 🎉 🎉 🎉 🎉

Issue #234, opened 2018-06-09
Server response body is swallowed if code is not OK


happycollision:
Pretty sure this was introduced at commit 7558b36

Since v4.0.0, if a server responds with something like this (with a non-ok response code):
```
{errors: [{
  status: 401,
  title: 'no user exists with that email and password combination',
}]}
```

The errors array disappears and the only thing available in the `catch` block of the call to `authenticate()` is a vanilla error where `message` is "Unauthorized".

Given the following:
```js
this.get('session').authenticate('authenticator:jwt', badCredentials)
  .catch(e => console.log(e));
```

3.x console:
```
{
  errors: [
    {
      status: 401,
      title: 'no user exists with that email and password combination',
    }
  ]
}
```

Post 4.0.0 console:
```
Error: Unauthorized
    at then.response (token.js:92)
    at tryCatcher (rsvp.js:215)
    at invokeCallback (rsvp.js:393)
    at publish (rsvp.js:379)
    at rsvp.js:14
    at invoke (backburner.js:205)
    at Queue.flush (backburner.js:125)
    at DeferredActionQueues.flush (backburner.js:278)
    at Backburner.end (backburner.js:410)
    at Backburner._boundAutorunEnd (backburner.js:372)
```

fenichelar:
@happycollision How does that look? In your example, the console output would be:

```
{
   "statusText": "Unauthorized",
   "status": 401,
   "json": {
      "errors": [
         {
            "status": 401,
            "title": "no user exists with that email and password combination"
         }
      ]
   }
}
```

happycollision:
That works *for me* just fine.

But are there scenarios where the server might not be sending JSON? Perhaps the error messages will be in plain text, etc. That discussion is up to you and your fellow collaborators.

In any case, thanks for that amazingly fast response!

fenichelar:
Yeah, I thought about. It was kind of messy so I figured it wasn't worth it but you are right. I added support for both: https://github.com/jpadilla/ember-simple-auth-token/pull/235/commits/32568baccda7014ba6081cf0bdb76c826b3601c5

JSON response:

```
{
  statusText: "Unauthorized",
   status: 401,
   text: "{"errors":[{"status":401,"title":"no user exists with that email and password combination"}]}",
   json: {
      "errors": [
         {
            "status": 401,
            "title": "no user exists with that email and password combination"
         }
      ]
   }
}
```

Text response:

```
{
  statusText: "Unauthorized",
[...]

PR #235, opened 2018-06-09
Improve fetch error handling, add ESA dependency, allow disabling session invalidation

fenichelar:
Fixes https://github.com/jpadilla/ember-simple-auth-token/issues/234
Fixes https://github.com/jpadilla/ember-simple-auth-token/issues/236
Fixes https://github.com/jpadilla/ember-simple-auth-token/issues/239

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,10 +15,9 @@ A demo is available [here][demo].
 
 ## Installation
 
-Before installing the Ember Simple Auth Token addon, [ember-simple-auth][ember-simple-auth] must be installed. Both addons can be installed with [Ember CLI][ember-cli] by running:
+Ember Simple Auth Token can be installed with [Ember CLI][ember-cli] by running:
 
 ```
-ember install ember-simple-auth
 ember install ember-simple-auth-token
 ```
 
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@ In addition to all the customization options available to the token authenticato
 ENV['ember-simple-auth-token'] = {
   tokenDataPropertyName: 'tokenData'; // Key in session to store token data
   refreshAccessTokens: true, // Enables access token refreshing
+  tokenExpirationInvalidateSession: true, // Enables session invalidation on token expiration
   serverTokenRefreshEndpoint: '/api/token-refresh/', // Server endpoint to send refresh request
   refreshTokenPropertyName: 'refresh_token', // Key in server response that contains the refresh token
   tokenExpireName: 'exp', // Field containing token expiration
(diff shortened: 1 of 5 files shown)


fenichelar:
Version 4.0.3. Changelog:

```
# Changelog

- Improve fetch error handling #234
- Add Ember Simple Auth to dependencies #236
- Allow disabling session invalidation on token expiration #239
```

happycollision:
I'm not entirely sure what special sauce is added to `ember install` vs `yarn add` or `npm install`, but is it possible that `ember-simple-auth` should be a *peer dependency* of ESAT?

Does changing the install instructions to simply 

```
ember install ember-simple-auth-token
```

 actually change how `ember-simple-auth` might be installed?

fenichelar:
See the change to `package.json`, `ember-simple-auth` was a development dependency and now it is a dependency. This means `ember install ember-simple-auth-token ` (or `npm install ember-simple-auth-token` or `yarn install ember-simple-auth-token`) will install `ember-simple-auth`. Therefore the install instructions were changed. This is the same as `ember-simple-auth-auth0`: https://github.com/auth0-community/ember-simple-auth-auth0/blob/develop/package.json#L84.

happycollision:
Right on. Like I said, I didn’t know if `ember install` worked any differently from `npm install` when it comes to dependencies.

jpadilla (commented):
@fenichelar I'm wondering if we should make less assumptions on these responses...

Was checking out what ESA does in their authenticators: https://github.com/simplabs/ember-simple-auth/blob/master/addon/authenticators/oauth2-password-grant.js

- They've included a `rejectWithResponse` property on the authenticator.
- https://github.com/simplabs/ember-simple-auth/blob/master/addon/authenticators/oauth2-password-grant.js#L326

Thoughts?

fenichelar:
@jpadilla Mmmm. What assumptions are you referring to? I think the result is pretty similar to oauth2. It returns JSON if possible and plain text if not.

It seems like oauth2 added `rejectWithResponse` for when you want to access the status code or status text:

> This is useful for cases when the backend provides additional context not available in the response body.

It seems simpler to just always include the status code and status text so that you don't need to override the callback. I think that is the only really difference unless I am missing something, which I totally could be?

fenichelar:
@jpadilla What do you think we should do?

jpadilla:
@fenichelar In that case could we include status code, status text, and response headers?

fenichelar:
@jpadilla 

> response headers

Done! Great suggestion. The headers are in a map because that is what `fetch` does: `response.headers.get('HEADER_NAME')`.

I also made `makeRequest` return all the response data for a successful return too. This makes the API more consistent and allows someone to easily change `authenticate` to support returning the token in a header or anything else like that.

jpadilla (approved):

PR #240, opened 2018-06-25
remove redundant try...catches

bekzod:


fenichelar (commented):
I believe the try/catch in `makeRequest` is required to achieve the desired behavior of handling both JSON and HTML errors.

bekzod:
@fenichelar I think I fixed that now :)

fenichelar:
@bekzod Can you please update to pass linter? Maybe something like this:

```js
return fetch(url, {
  method: 'POST',
  headers: assign({
    'Accept': 'application/json',
    'Content-Type': 'application/json'
  }, headers),
  body: JSON.stringify(data)
}).then(response => {
  const res = {
    statusText: response.statusText,
    status: response.status,
    headers: response.headers
  };
  response.text().then(text => {
    res.text = text;
    try {
      res.json = JSON.parse(text);
    } catch (e) {
      return reject(res);
    }
    if (response.ok) {
      return resolve(res);
    } else {
      return reject(res);
    }
  }).catch(() => {
    return reject(res);
  });
}).catch(error => {
  return reject(error);
});
```

bekzod:
@fenichelar done

fenichelar (approved):

PR #243, opened 2018-06-26
cleanup promises

bekzod:
given that `this.makeRequest` is already wrapped in `RSVP/Ember Promise` there is no need wrap it again with `RSVP` Promise

### DIFF
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -125,11 +125,8 @@ export default TokenAuthenticator.extend({
                                  otherwise
   */
   authenticate(credentials, headers) {
-    return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


addon/authenticators/jwt.js
```diff
@@ -244,7 +239,7 @@ export default TokenAuthenticator.extend({
     delete this._refreshTokenTimeout;
     cancel(this._tokenExpirationTimeout);
     delete this._tokenExpirationTimeout;
-    return new resolve();
```
bekzod, line 247: this doesn't need `new`, it already returns new instance of Promise

fenichelar (approved):

fenichelar:
Awesome! :)

PR #246, opened 2018-08-08
 Support multiple ember fetch versions

fenichelar:
I am getting duplicate copies of ember fetch in my applications. We should keep this the same as ember-simple-auth to prevent this.

```
# Changelog

-  Add support for multiple ember-fetch versions #246
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,36 +5,44 @@
   "requires": true,
   "dependencies": {
     "@ember/ordered-set": {
-      "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


jpadilla (approved):

jpadilla:
@fenichelar good catch!

fenichelar:
@jpadilla It would be awesome to get this released on npm when you have a chance.

jpadilla:
@fenichelar done!

Issue #247, opened 2018-08-13
[Fastboot] atob() global function is not available in Node.js environment


musaffa:
`atob` global function [used](https://github.com/jpadilla/ember-simple-auth-token/blob/5102353fc2178582c3f9aaa86fa12995a6072603/addon/authenticators/jwt.js#L221) in jwt authenticator is only a browser specific implementation. We get a reference error when calling `atob` in Fastboot's Node.js environment.

A possible solution is to use an isomorphic implementation of `atob`. For example: [abab](https://github.com/jsdom/abab).

fenichelar:
@musaffa Mmmm, I'm not sure I understand the use case. If you are calling `getTokenData` (either directly or indirectly) on the server, what session data would be used?

musaffa:
Cookie store. See [Fastboot](https://github.com/simplabs/ember-simple-auth#fastboot) section of Ember Simple Auth's readme.

fenichelar:
@musaffa Does #248 fix your issue?

musaffa:
No it doesn't. I've tested this commit in a Fastboot application. I've also shared my observation on that PR.

PR #248, opened 2018-08-14
User base64 instead of atob for fastboot support

fenichelar:
Fixes https://github.com/jpadilla/ember-simple-auth-token/issues/247

`window.base64` from https://github.com/simplabs/ember-simple-auth/blob/master/addon/authenticators/oauth2-password-grant.js#L122

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ Ember Simple Auth Token can be installed with [Ember CLI][ember-cli] by running:
 ember install ember-simple-auth-token
 ```
 
+If using FastBoot and the JWT authenticator, `buffer` must be added to your `fastbootDependencies`.
+
 ## Setup
 
 ### Authenticator
--- a/addon/authenticators/jwt.js
+++ b/addon/authenticators/jwt.js
@@ -162,7 +162,7 @@ export default TokenAuthenticator.extend({
           delete this._refreshTokenTimeout;
           this._refreshTokenTimeout = later(this, this.refreshAccessToken, refreshToken, wait);
         } else if (expiresAt > now) {
-          throw new Error('refreshLeeway is too large which is preventing token refresh');
+          throw new Error('refreshLeeway is too large which is preventing token refresh.');
         }
       }
     }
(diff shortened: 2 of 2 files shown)


fenichelar:
@jpadilla Yes, but it is probably not a good idea to depend on that undocumented behavior. I will reproduce internally. Good point!

fenichelar:
@jpadilla I think that is better.

jpadilla:
@musaffa any chance you could test this out and see if it resolves your original issue?

jpadilla (dismissed):

fenichelar:
Yes, `buffer` must be added to `fastbootDependencies`. We should add some documentation about this.

------------------------------------------------

Using the `FastBoot` should be no different than `require`:

> From your Ember.js app, you can run FastBoot.require() to require a package. This is identical to the CommonJS require except it checks all requests against the whitelist first.

I'm hesitant to use the `FastBoot` global in this library.

------------------------------------------------

> A simple (atob) conditional check doesn't work.

What does it do?

------------------------------------------------

> Unable to require module 'buffer' because it was not in the whitelist

That seems like a good error message to me... however, it will not be displayed if the `FastBoot` global is not used. So custom error message makes sense.

------------------------------------------------

Update incoming.

musaffa:
> > A simple (atob) conditional check doesn't work.
> 
> What does it do?

`(typeof atob !== 'undefined')` is the right way to check for the presence of atob. The simple check currently present in the PR doesn't work.

musaffa:
IMHO you should test the changes in a Fastboot environment.

fenichelar:
@musaffa I did... and it worked fine. I did add buffer to the fastbootDependencies.

fenichelar:
```js
const decode = str => {
  if (typeof atob === 'function') {
    return atob(str);
  } else if (typeof FastBoot === 'object') {
    try {
      const buffer = FastBoot.require('buffer');
      return buffer.Buffer.from(str, 'base64').toString('utf-8');
    } catch (err) {
      throw new Error('buffer must be available for decoding base64 strings in FastBoot. Make sure to add buffer to your fastbootDependencies.');
    }
  } else {
    throw new Error('Neither atob nor the FastBoot global are avaialble. Unable to decode base64 strings.');
  }
};
```

I think that should be good.

musaffa:
LGTM

fenichelar:
@jpadilla Thoughts on this?

jpadilla (approved):

PR #259, opened 2019-01-02
Fix Return For Token Authenticator

jherdman:
I pinpointed an issue I was experiencing with the Token Authenticator
and it _always_ returning `undefined`. This seemed to be something to do
with unwrapping Promise values. The `makeRequest` method was
unnecessarily wrapping `fetch`. In removing this wrapping and leaning
more on built in Promise semantics the problem went away.

### DIFF
--- a/addon/authenticators/token.js
+++ b/addon/authenticators/token.js
@@ -1,7 +1,7 @@
 import EmberObject from '@ember/object';
 import fetch from 'fetch';
 import { assign } from '@ember/polyfills';
-import { Promise, resolve } from 'rsvp';
+import { Promise, reject, resolve } from 'rsvp';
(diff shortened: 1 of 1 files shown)


jherdman:
To clarify, `this.session.authenticate(...)` still returns `undefined`, but now I can actually access the session data.

jpadilla:
@fenichelar thoughts on this, at a glance it LGTM

fenichelar (approved):

fenichelar:
@jherdman Nice clean up!

@jpadilla Agreed.

I'm not sure I understand how this changes anything with regards to accessing the session data and I am able to access it as it currently stands. But this is definitely a good improvement from a simplification standpoint.

PR #260, opened 2019-02-21
config/environment.js semicolon to comma update.

yusufsagdic:
It return build error when I added to environment.js because of semicolon.


fenichelar (approved):

fenichelar:
Thanks for the fix @yusufsagdic!

PR #270, opened 2020-02-24
Support ember-simple-auth v2.1

fenichelar:


fenichelar:
Travis is driving me up a wall! I can't seem to get it to stop using npm version v3.5.4...

tben:
Change "npm install -g npm@6" to "npm install npm@6" in travis file. It's due to that fact that ember-cli-release requires a really old version of npm. Might be worth moving to "ember-cli-release-tag" which is a more maintained version.

fenichelar:
@tben Thank you!!

I am going to clean up my commit log (rewrite history) before we try and get this merged.

fenichelar:
Changelog:

```
# Changelog

- Add testing for Node.js v8 and v10 in Travis CI
- Add testing for Ember LTS 3.4, 3.8, 3.12, 3.16
- Add support for ember-simple-auth v2.1 and v3
- Remove ember-cli-release
- Update ember-try
- Replace deprecated merge with assign
```

fenichelar:
@jpadilla 

Do we want to drop Ember 2.16? It is no longer supported. If yes, I think we should remove it before we merge this. 

Other than that, thoughts?

jpadilla:
> Do we want to drop Ember 2.16? It is no longer supported. If yes, I think we should remove it before we merge this.

@fenichelar sounds good to me, thanks for working on this! 🎉

jpadilla (dismissed):

package.json
```diff
@@ -25,7 +25,7 @@
     "ember-cli-babel": "^6.6.0",
     "ember-get-config": "^0.2.4",
     "ember-fetch": "^4.0.0 || ^5.0.0 || ^6.0.0",
-    "ember-simple-auth": "^1.6.0"
+    "ember-simple-auth": "^1.6.0 || ^2.1.0"
```
jelhan: I think we should also add support for `ember-simple-auth@^3.0.0`:

```suggestion
    "ember-simple-auth": "^1.6.0 || ^2.1.0 || ^3.0.0"
```

package.json
```diff
@@ -41,7 +41,7 @@
     "ember-cli-htmlbars-inline-precompile": "^1.0.0",
     "ember-cli-inject-live-reload": "^1.4.1",
     "ember-cli-qunit": "^4.1.1",
-    "ember-cli-release": "^1.0.0-beta.2",
+    "ember-cli-release-tag": "^0.1.5",
```
jelhan: I would recommend to change to [`release-it`](https://github.com/release-it/release-it#readme). Much more powerful and better maintained alternative.

It's often used together with [`lerna-changelog`](https://github.com/lerna/lerna-changelog), which automates the creation of a changelog based on GitHub pull requests and labels used for them. Setup could be automated with [`create-rwjblue-release-it-setup`](https://github.com/rwjblue/create-rwjblue-release-it-setup).

package.json
```diff
@@ -77,13 +77,13 @@
     "authorization"
   ],
   "engines": {
-    "node": ">= 6.*"
+    "node": ">= 8.*"
```
jelhan: Node 8 is already EOL. `ember-simple-auth@^3.0.0` also requires Node >= 10. Would recommend to drop support for node 8 as well to avoid another major release.

```suggestion
    "node": ">= 10.*"
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ember-simple-auth-token",
-  "version": "4.0.7",
+  "version": "4.0.8",
```
jelhan, line 3: Dropping support for node 6 is a breaking change and requires a major version release. But I think versions are not updated as part of pull requests at all, are they?

jelhan (commented):
Thanks a lot for working on this. I also need it for one of my projects. But would love to upgrade to `ember-simple-auth@v3.0.0` directly. Added some review comments that would allow me to do that.

Not sure if supporting multiple versions of `ember-simple-auth` is really necessary at all. But if we do it, we should run the tests against all supported version.

package.json
tben: I somewhat disagree, I believe there should be at least one release with support for node 8 and ember-simple-auth v2 for those that haven't got the latest version or are unable to.

tben: I somewhat disagree, I believe there should be at least one release with support for node 8 and ember-simple-auth v2 for those that haven't got the latest version or are unable to.

package.json
fenichelar: I'm a believer in making releases as small as possible. For now, we let's leave this as is.

package.json
fenichelar: I have added support for ember-simple-auth v3 but not done it in such a way that node v10 is required. In fact, I have also reverted a previous change so that node v6 is also still supported.

package.json
fenichelar: I have reverted this to support node v6.

package.json
fenichelar: Good point. Node v6 is now still supported. There is no need for this library to force a specific version.

fenichelar:
@tben and @jelhan thoughts?

@jpadilla can you please review when possible?

jelhan:
I'm still worried about the tests.

`package.json` allows three major versions of `ember-simple-auth`:

```json
{
  "dependencies": {
    "ember-simple-auth": "^1.6.0 || ^2.1.0 || ^3.0.0" 
  }
}
```

But the tests are only run against the version of `ember-simple-auth` currently locked down in `package-lock.json`, which is `2.1.1`. The _Floating Dependencies_ test scenario which is included in latest Ember CLI blueprints is not used in this addon. But it wouldn't help a lot as it would be unpredictable, which major version is picked.

Supporting multiple versions of a dependency increases testing complexity a lot. Therefor I tend to avoid it. But if you want to go that path, the tests should also cover all possible combinations. Or at least cover one common combination per scenario. Maybe using `ember-simple-auth@v1` with `node@v6` as default for all tests and testing additionally these two scenarios:

- `ember-simple-auth@v2` with `node@v8` and `ember-source@2.16`
- `ember-simple-auth@v3` with `node@v10` and `ember-source@3.0`

fenichelar:
@jelhan I think writing tests for every version is not practical. My goal here is to just offer one last version with support for all 3 versions of node and ember-simple-auth. Then make a new major version here with more limited support.

I am still debating the majors, options:

1. Two majors, one for node 8 and ember-simple-auth 2 and one for node 10 and ember-simple-auth 3
2. One major, node 10 and ember-simple-auth 3

I'm leaning towards the first but we would again be lacking tests. But tests can be added if needed, dropping support is permanent.


What do you think of that plan?

jelhan:
> I think writing tests for every version is not practical.

It's not that much work. Ember-try provides nearly everything that we need out of the box. Had some time. Please find the pull request adding missing test coverage here: https://github.com/fenichelar/ember-simple-auth-token/pull/1

> Then make a new major version here with more limited support.

As we now have a version that support all three version of ember-simple-auth and has test coverage I don't see a need for a major release. We already invested all the time needed. I'm still not convinced that spending all that time on this one was a good decision but we should now also earn the benefits. :smile: 

Only one concern left from my side:

This pull requests drops support for node 6. This is a breaking change and would require a major version. The only reason it does so is adding `ember-cli-release-tag`. I'm not sure if a change to the internal release process justifies a new major version.

https://github.com/fenichelar/ember-simple-auth-token/pull/2 adds node 6 support back and adjusts the test coverage accordingly.

fenichelar:
@jelhan Travis does not appear to be using Node v10 for the test case

jelhan:
> @jelhan Travis does not appear to be using Node v10 for the test case

Could you elaborate? Tests are passing and it correctly reports to install and use node 10: https://travis-ci.org/jpadilla/ember-simple-auth-token/jobs/658757270#L485-L491

fenichelar:
@jelhan You are correct. It installs Node v8, prints the version, and then installs Node v10. A little confusing but it does appear to be working.

jelhan (dismissed):
This is ready to be merged in @jpadilla. It should not be a breaking change beside for removing tools that help with releasing. But maybe that's okay for now? Or should we try to add `ember-cli-release` back in?

fenichelar:
@jelhan @jpadilla I'm going to add a few more tests for completeness. I'll let you know once ready.

fenichelar:
@jelhan Assuming all of the tests pass, what do you think about everything now?

BTW, I had no idea how powerful ember-try was. Thanks for teaching me something new!!

I think after this we should consider releasing v5 which would drop support for Node v6 and Node v8. I'd like to continue to support all versions of ember and ember-simple-auth as long as practical. Thoughts?

fenichelar:
@jpadilla thoughts?

jelhan:
> @jelhan Assuming all of the tests pass, what do you think about everything now?

To be honest I'm not sure if we need such a complete matrix to cover all possible combinations. But as this library is not changing that often it should not eat up to much CI resources. So I don't think that will be an actual problem. Would be great if this one could be merged and released.

fenichelar:
@jelhan I went a little overboard because I was planning on dropping support for Node v6 and Node v8 immediately after this is merged. What do you think about that? Looking for feedback.

jelhan:
> I was planning on dropping support for Node v6 and Node v8 immediately after this is merged.

Makes sense to me. Should also consider dropping support for `ember-simple-auth@v1` and `ember-simple-auth@v2` at least as soon as it adds any additional maintenance costs.

fenichelar:
@jelhan 

If you add your review, I can merge this.

I was thinking about leaving support for all three version of ember-simple-auth. For me at least, upgrading Node is easy, rarely an issue. Upgrading ember and other ember libraries can be a big project with lots of breaking changes. I'm using Node v10 but still using ember 2.x and ember-simple-auth v2 in a lot of places because of this. I imagine I'm not the only one in this position. Also, so far, supporting all version of ember-simple-auth has been trivial as there have been no breaking changes that affect this library.

fenichelar:
@jpadilla release-it v9.8.3 supports Node v6. This could be an alternative to ember-cli-release. But the tentative plan is to drop Node v6 and Node v8 in the next release so we may want to just add in ember-cli-release-tag or the latest release-it at that time.

jelhan (approved):
Looks good to me but I don't have write access to this repo. Need @jpadilla for merging and releasing.

fenichelar:
@jelhan Ahh, sorry. Yes, we need @jpadilla to review.

Changelog:
```
# Changelog

- Add testing for Node.js v8 and v10 in Travis CI
- Add testing for Ember LTS 3.4, 3.8, 3.12, 3.16
- Add support for ember-simple-auth v2 and v3
- Remove ember-cli-release
- Update ember-try
- Replace deprecated merge with assign
```

jpadilla:
@fenichelar awesome work, thanks! switching over to ember-cli-release-tag doesn't seem like a bad idea.

jpadilla (approved):

jpadilla:
@fenichelar What's your npm username? I can add you as a maintainer there too

fenichelar:
It was mostly @jelhan!

My NPM username is `fenichelar`

jpadilla:
Thank you too @jelhan 🎉

fenichelar:
@jpadilla I don't have `npm publish` permissions. Can you please publish this, v4.0.8.

fenichelar:
@jpadilla Thanks. Working now.

PR #274, opened 2020-04-26
Add token adapter mixin, fixes #273

### LINKED ISSUE #273: authorize assertion removed from Ember Simple Auth 2.1.1.
According to the ember-simple-auth repo the authorize assertion should not be used any more:
https://github.com/simplabs/ember-simple-auth/pull/2010

The authorize assertian has been removed from Ember Simple Auth 2.1.1. This breaks the following code:
https://github.com/jpadilla/ember-simple-auth-token/blob/facc3bea47196abc553d0248e3aca8c43254440d/addon/mixins/token-authorizer.js#L39-L48

I'll work on a PR to resolve this if is hasn't been fixed already and I'm just looking in the wrong place.

fenichelar:


fenichelar:
Fixes #273

fenichelar:
v5.1.0
```
# Changelog

- Add `token-adapter` mixin, fixes #273
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ember-simple-auth-token",
-  "version": "5.0.0",
+  "version": "5.0.1",
```
jpadilla: Should we bump minor instead of patch?

tests/unit/mixins/token-adapter-test.js
```diff
@@ -0,0 +1,11 @@
+import EmberObject from '@ember/object';
+import TokenAdapterMixin from 'ember-simple-auth-token/mixins/token-adapter';
+import { module, test } from 'qunit';
+
+module('Unit | Mixin | token-adapter', function() {
+  test('it works', function (assert) {
+    const TokenAdapterObject = EmberObject.extend(TokenAdapterMixin);
+    const subject = TokenAdapterObject.create();
+    assert.ok(subject);
+  });
```
jpadilla, line 10: could we add a simple test for the `headers` computed property?

jpadilla (changes requested):

package.json
fenichelar: That makes sense. I will update it to 5.1.0

tests/unit/mixins/token-adapter-test.js
fenichelar: Will do

package.json
fenichelar: This has been updated.

tests/unit/mixins/token-adapter-test.js
fenichelar: I have having some difficulties. Have you every written unit tests for a computed property in a mixin?

fenichelar: I gave up

jpadilla (approved):

PR #281, opened 2020-07-29
Update readme and packages

### LINKED ISSUE #279: Adapter vs Authorizer mixin, which one to use?
Hi there, 
From looking at the README, it is not clear to me whether I should use the new adapter or authorizer mixin.

They seem to be serving the same purpose.

> In order to send the token with all API requests made to the server, the token adapter mixin should be used:
> ...
> The mixin will add the header to each API request:
> ...

vs

> In order to send the token with all API requests made to the server, the token authorizer mixin should be used:
> ...
> The mixin will add the header to each API request:
> ...

When/where would I use one over the other?

Thank you.

fenichelar:
Fixes https://github.com/jpadilla/ember-simple-auth-token/issues/279

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,7 @@
 
 This is Ember addon is an extension to the Ember Simple Auth library that provides a basic token authenticator, a JSON Web Tokens token authenticator with automatic refresh capability, and an authorizer mixin. You can find more about why JSON Web Tokens are so awesome in [this article][medium-jwt].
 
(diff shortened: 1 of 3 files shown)


fenichelar:
v5.1.1
```
# Changelog

- Update readme to improve mixin documentation
- Update packages to address security vulnerabilities
```

jpadilla (approved):
✨ thanks!

PR #291, opened 2021-01-17
Add Ember v3.20.x tests

fenichelar:
v5.2.1
```
# Changelog

- Add Ember v3.20.x tests
- Add Ember v3.24.x tests
- Update packages
```


jpadilla (approved):
Should we drop support already for anything ember<3.0?

fenichelar:
@jpadilla I personally still use ember 2.x in a number of applications so I would prefer to keep it. The only downside is that running all the test scenarios takes awhile.

PR #293, opened 2021-01-18
Move ESA to peerDependency and update tests. 

BryanCrotaz:
Moved ESA to peerDependency.
Use yarn for tests as maintainer says tests run slow. Yarn is 3x faster in many scenarios.
Add test scenarios for canary of ESA


fenichelar:
See https://github.com/jpadilla/ember-simple-auth-token/issues/290#issuecomment-762378690

This does not solve the problem. It just shift it somewhere else.

PR #295, opened 2021-01-18
Add install note to readme

### LINKED ISSUE #290: Possible to get a beta release that is is sync with ember-simple-auth 3.1.0-beta.0?
Hi there,

I was trying to get my Ember app up to date (v3.24 as of this writing), which involves getting rid of older ways of doing things, such as using mixins. Well, in order to do that, I had to upgrade ember-simple-auth to 3.1.0-beta.0 (or .1). However, when I run my app, ember-simple-auth-token has a dependency on ESA v3.0.1 or so, which seems to override the ESA version for my whole app, so the new API does not work.

This might be a big ask since I am not familiar with this project and do not know how much is involved, but would it be possible to get a beta release of ESA-token that supports the latest betas of ESA? I realize there are potentiall any number of issues that need to be resolved to support the new API, but at least one thing is positive; Editing the node_modules/ember-simple-auth-token/package.json in my project to refer to ESA 3.1.0-beta.0 seems to work fine. I just get two deprecations, that is all.

I would be happy to contribute towards this, but I am at a very basic level with Ember so I might now have a lot to bring to the table.

Thanks for any insights!

fenichelar:
Fixes https://github.com/jpadilla/ember-simple-auth-token/issues/290

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,8 @@ ember install ember-simple-auth-token
 
 If using FastBoot, `ember-fetch` must be installed as a direct dependency and `node-fetch` must be added to your `fastbootDependencies`. If using FastBoot and the JWT authenticator, `node-fetch` and `buffer` must be added to you `fastbootDependencies`.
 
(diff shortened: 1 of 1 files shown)


fenichelar:
Hey @jpadilla. When you get a chance, can you please review?

jpadilla (approved):

Issue #302, opened 2021-09-18
Addon incompatible with latest version of ember-simple-auth


maxymczech:
This pull request has removed `setup-session-service`: https://github.com/simplabs/ember-simple-auth/pull/2315

Rendering `ember-simple-auth-token` incompatible.

maxymczech:
I apologize, after further investigation I can elaborate. I am using new `embroider` build system with `webpack`. After updating `ember-simple-auth` to version `4.0.0` the app I am working on was broken. The reason for this, as it seems, is simply that you don't have `ember-simple-auth@4.0.0` in your dependencies list:

`"ember-simple-auth": "^1.6.0 || ^2.0.0 || ^3.0.0"`

The app works fine if I revert back to classic build, and it also works if I try and add `^4.0.0` to that dependency version list. I hope this could be just as easy as that.

fenichelar:
@maxymczech Yes, `ember-simple-auth@v4` isn't supported yet. I hope to work on this in the next few weeks.

fenichelar:
`ember-simple-auth-token@v5.3.0` supports `ember-simple-auth@v4`

PR #303, opened 2021-09-18
Add ember-simple-auth@4.0.0 to dependency list

maxymczech:
Solves #302


fenichelar:
@maxymczech There is a fair amount of additional work that needs to be done for this with tests. I have started working on it. I hope to have everything completed in the next few weeks.

maxymczech:
Thank you very much. Let me know if I can be of help.

